PR #3, opened 2016-03-03
Update aerospike enterprise logrotate conf

roa:
- add missingok flag
- test if pid file exists before reloading aerospike


ferki:
LGTM :+1:

PR #5, opened 2016-03-14
Add zabbix-3.0.0 ebuild

roa:


patricklauer:
(1) why not 3.0.1?
(2) we can upstream it directly if you have confidence that it works

( see also https://bugs.gentoo.org/show_bug.cgi?id=574858 )

roa:
a) 3.0.1 was not around when i started working on it.
b) i created it for evaluation purposes and i am not confident, that it adds all new 3.x features / that it works on all machines :)

i believe, if someone is already working on it, we can wait until it hits portage.

ferki:
I fixed merge conflicts, updated commit message and added `metadata.xml`.

PR #8, opened 2016-03-30
Update grafana ebuild

ferki:
- rename to www-apps/grafana-bin
 - use upstream's sample config
 - slight deps update


patricklauer:
lgtm

PR #15, opened 2016-05-02
Fix dev-perl/YAML dependencies

roa:
dev-perl/yaml was renamed to dev-perl/YAML
=> updated the DEPEND and RDEPEND sections


ferki:
:+1:

PR #27, opened 2016-09-14
net-analyzer/prometheus: Add ebuild

roa:
Package-Manager: portage-2.2.28


net-analyzer/prometheus/prometheus-1.1.2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+# $Id$
+
+EAPI=5
+
+inherit golang-base user
+
+DESCRIPTION="The Prometheus monitoring system and time series database"
+HOMEPAGE="http://prometheus.io"
+EGO_PN="github.com/${PN}/${PN}"
+SRC_URI="https://${EGO_PN}/archive/v${PV}.tar.gz"
```
patricklauer: please use a SRC_URI arrow to rename it:
SRC_URI="https://${EGO_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"

PR #31, opened 2016-10-26
app-misc/flink-bin: Initial ebuild

roa:
Package-Manager: portage-2.3.0


app-misc/flink-bin/flink-bin-1.1.3.ebuild
```diff
@@ -0,0 +1,46 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	|| ( virtual/jre:1.8 )
+"
+DEPEND="${RDEPEND}"
+
+S="${WORKDIR}/${MY_PN}-${PV}"
+INSTALL_DIR="/opt/${MY_PN}-${PV}"
```
patricklauer: Since it's not slotted I'd not use PV in the install dir, because every upgrade will shift the path names

app-misc/flink-bin/flink-bin-1.1.3.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+pkg_setup() {
+	enewgroup flink
+	enewuser flink -1 /bin/bash /var/lib/flink flink
+}
+
+src_install() {
+	keepdir /var/lib/flink
```
patricklauer: No init script?

PR #36, opened 2016-11-16
sys-kernel: New version of buildkernel and gentoo-kernel 

patricklauer:
With linux-firmware bundled in


patricklauer:
This is partially obsoleted by 4.8.9

PR #42, opened 2017-01-23
sys-kernel: Add kernel 4.9.5

patricklauer:
New LTS, new fun.


patricklauer:
Performance regression, try 4.10

PR #45, opened 2017-02-07
dev-db/aerospike-server-enterprise: Bump to 3.11.0.2

patricklauer:
Needs some shakedown testing first, but files seem to be in the right places.


patricklauer:
Obsoleted for 3.12

PR #47, opened 2017-02-20
sys-kernel: Bump to 4.10.0

patricklauer:
Package-Manager: Portage-2.3.3, Repoman-2.3.1


alip:
lgtm, builds and installs fine in an updated stage4 chroot.

PR #49, opened 2017-03-03
Revbump to logstash-1.5.4-r2 (updated init)

alip:
Updated init file provides basic support to configure nice level.
Required for adjust/infrastructure#1517


patricklauer:
if you read through `/etc/rc.conf` ...
```
# Some daemons are started and stopped via start-stop-daemon.
# We can set some things on a per service basis, like the nicelevel.
#SSD_NICELEVEL="-19"
```
We could just use that in `/etc/conf.d/logstash` without having to touch anything else.

alip:
That I did not know, I'll amend :)

alip:
Even better is we can drop this revbump and upgrade infrastructure.git

PR #53, opened 2017-03-29
www-apps/grafana-bin: bump version and fix issues

ferki:
This ebuild bumps `www-apps/grafana-bin` to latest released version, 4.2.0.

It also addresses current open issues in Gentoo's upstream tracker, namely:
 - [bug 601282](https://bugs.gentoo.org/show_bug.cgi?id=601282): missing systemd service file
  Systemd service file is taken from upstream's deb package, and then adapted to Gentoo.

 - [bug 608774](https://bugs.gentoo.org/show_bug.cgi?id=551938): directory/ownership/permission fixes
   - Logrotate config was not included as grafana has built-in logrotation capabilites, configurable via its own `grafana.ini` file, and thus there's a risk of conflict if more than one rotation method is in place. It's still possible for users to disable the built-in rotation, and roll out their own preferred logrotate configuration.
   - Making the various directories world-readable was not included. Instead unprivileged users wishing to access e.g. logs of grafana, can be added to the `grafana` group which already should have proper permissions in place.
   - `plugins` and `sessions` directories are also taken care of on top of the suggested `dashboards` directory. It is made sure now they are not only being created and set to proper ownership/permissions upon first-time installation, but also if they were created already with different settings.

On top of that, it also fixes some QA notices.

Please review and let me know if further modifications are needed. I'd like to squash the commits together before merging, but feel free to merge like this if you prefer that.


ferki:
The same changes has been sent to Gentoo upstream via [bug 614292](https://bugs.gentoo.org/show_bug.cgi?id=614292).

ferki:
@patricklauer: asked a review from you in hopes you can get this bumped in gentoo upstream as well. Feel free to reassign the review to someone else if that would lead to faster merge in this overlay :)

patricklauer:
it's in repo/gentoo

PR #55, opened 2017-03-30
app-misc/karmabot-bin: add missing metadata.xml

ferki:
Lack of `metadata.xml` breaks tests not just for master branch, but also for all branches trying to fork from there (e.g. any new work), or trying to rebase on top of current master (e.g. in-progress work).

Please use `repoman commit` for committing your ebuild work to prevent such issues, and also please respect automated test results in case you forgot to use `repoman commit` :)


jolenzy:
:+1: 
Thanks

PR #56, opened 2017-04-03
app-admin/schaufel: add live ebuild

roa:
Package-Manager: Portage-2.3.3, Repoman-2.3.1


app-admin/schaufel/schaufel-9999.ebuild
```diff
@@ -0,0 +1,36 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+inherit user
+
+DESCRIPTION="shovel data around"
+HOMEPAGE="https://github.com/adjust/schaufel"
+SRC_URI="https://github.com/adjust/schaufel/archive/${PN}-9999.tar.gz"
```
patricklauer: This will fail checksums on each upstream commit added ... better to use the `git-r3` eclass, I think

PR #60, opened 2017-05-02
sys-kernel: Bump to kernel 4.11

patricklauer:
New kernel version.
Adds bcache support once more.
Switch default IO scheduler to deadline


jolenzy:
Looks good :+1: 
I'm curious to see the impact of deadline scheduler.

patricklauer:
On the elastic indexers it gave us a noticeable boost (~20% under heavy load), I expect it to be less for other machines (and we can always manually override if required)

jolenzy:
Okay, that sounds great :)

patricklauer:
There's still some weirdness that I haven't had time to debug, with 4.12 on the horizon I don't think we'll fix this. Closing

PR #62, opened 2017-05-22
Ebuildify wkhtmltopdf with qt patch

alip:


alip:
repoman is almost completely happy now.

wiktor-b:
Why are we updating copyright years of unchanged ebuilds?

alip:
Because repoman moans for `Invalid copright year` which in turn makes travis fail. As far as I know, Gentoo never had a policy to depend on last change to update copyright years anyway (@patricklauer would know better, please correct me if I'm wrong) so I do not think this is a big issue.

PR #68, opened 2017-06-20
Revbump to add patch & USE to opt out of ltree extension

alip:


dev-db/postgresql/files/postgresql-9.6.3-wo-ltree.patch
```diff
@@ -0,0 +1,68 @@
+Source: Written by Ali Polatel <alip@exherbo.org>
+Upstream: Not yet (TODO: patch must be updated for git://)
+Reason: Make the ltree extension build optional via --without-tree
```
patricklauer: --without-tree looks like it should be --without-ltree ?

dev-db/postgresql/postgresql-9.6.3-r101.ebuild
```diff
@@ -23,7 +23,7 @@ HOMEPAGE="http://www.postgresql.org/"
 
 LINGUAS="af cs de en es fa fr hr hu it ko nb pl pt_BR ro ru sk sl sv tr
 		 zh_CN zh_TW"
-IUSE="doc kerberos kernel_linux ldap libressl nls pam perl -pg_legacytimestamp python
+IUSE="doc kerberos kernel_linux ldap libressl +ltree nls pam perl -pg_legacytimestamp python
```
patricklauer: I think for us it would make more sense to default to ltree disabled.

alip:
@patricklauer, the last commit fixes both of your concerns above.

alip:
@patricklauer: ping ^^ :)

alip:
I have added ltree USE flag for all latest versions in SLOTs 9.{4,5,6} and 10_beta2. The changeset is minimal and described in the latest commit above, namely 27633b0 so adding in new minor postgresql versions is trivial.
@patricklauer: We can consider pushing this to gentoo-x86 as well.

patricklauer:
I think the useflag should be default-enabled - IUSE="+ltree"

There's about zero users outside of us that benefit from it, so I don't think we want to push it upstream.

alip:
> I think the useflag should be default-enabled - IUSE="+ltree"

Quoting your last review which is not available here due to my last rebase I guess:
> @patricklauer commented on this pull request.
> In dev-db/postgresql/postgresql-9.6.3-r101.ebuild:
> ...
> -IUSE="doc kerberos kernel_linux ldap libressl nls pam perl -pg_legacytimestamp python
> +IUSE="doc kerberos kernel_linux ldap libressl +ltree nls pam perl -pg_legacytimestamp python
>
> I think for us it would make more sense to default to ltree disabled.

I am confused :) To me default-disabled makes much more sense taking into account that we won't upstream this and the only remaining use case is to always disable ltree USE flag. 

> There's about zero users outside of us that benefit from it, so I don't think we want to push it upstream.

Ok, no upstream it is.

patricklauer:
Eh yes. Let's stick with default-disabled for our ebuilds.

PR #69, opened 2017-06-28
sys-kernel/{buildkernel,gentoo-{sources,kernel}}: Bump

patricklauer:
Make AHCI a builtin to work around race conditions in device init


patricklauer:
Obsolete.

PR #79, opened 2017-08-15
Initial ebuild for net-misc/google-cloud-sdk

alip:
Based in part upon google-cloud-sdk-138.0.0.ebuild of coreos-overlay,
with the following changes:
1. Updated to current version
2. Fixes & tweaks to work around python eclass.
3. Install under /opt rather than /usr/lib
4. Remove some more bundled python libraries (still more left as TODO)

Ref: adjust/infrastructure#2317


jolenzy:
As far as I can tell, looks OK :+1:

PR #82, opened 2017-08-24
Adding PGObject-Util-PGConfig to overlay.

einhverfr:
This is a basic ebuild generated with g-cpan and reviewed/edited for clarity.


dev-perl/PGObject-Util-PGConfig/PGObject-Util-PGConfig-0.02.ebuild
```diff
@@ -0,0 +1,22 @@
...
+MODULE_VERSION="0.02"
+
+
+inherit perl-module
+
+DESCRIPTION="Postgres Configuration Management"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="alpha amd64 amd64-fbsd arm arm64 hppa ia64 m68k mips nios2 ppc ppc64 riscv s390 sh sparc sparc-fbsd x86 x86-fbsd   ppc-aix amd64-linux arm-linux arm64-linux ppc64-linux x86-linux ppc-macos x86-macos x64-macos m68k-mint sparc-solaris sparc64-solaris x64-solaris x86-solaris x86-winnt x64-cygwin x86-cygwin"
```
patricklauer: Since we've only tested on amd64 that's the only one you should add. Also, since it's a new ebuild it should be ~amd64 by convention

einhverfr: Nice to see what needs to change from the tools for auto conversion.  Making this change quickly.

PR #83, opened 2017-08-24
Adding Postgres exporter for Prometheus to repo

einhverfr:
Old ebuild used without changes, found on gpo.zugaina.org.


net-analyzer/prometheus-postgres_exporter/prometheus-postgres_exporter-0.2.2.ebuild
```diff
@@ -0,0 +1,59 @@
...
+EXPORTER_USER=prometheus-exporter
+
+MY_P="${P##prometheus-}"
+S="${WORKDIR}/${MY_P}/src/${EGO_PN}"
+
+pkg_setup() {
+	enewuser "${EXPORTER_USER}" -1 -1 -1
+}
+
+src_unpack() {
```
einhverfr: Maybe I will try with that.  Better to do it right the first time.

PR #84, opened 2017-08-30
Perl replication management classes

einhverfr:
This adds the modules I wrote in the PGObject::Util::Replication namespace and will have to be used along with repmgr.


dev-perl/PGObject-Util-Replication-Master/PGObject-Util-Replication-Master-0.01.ebuild
```diff
@@ -10,6 +10,7 @@ inherit perl-module
 
 DESCRIPTION="Manage Database Masters"
 SRC_URI="mirror://cpan/authors/id/E/EI/EINHVERFR/${PN}-v${PV}.tar.gz"
+S=/var/tmp/portage/dev-perl/PGObject-Util-Replication-Master-0.01/work/PGObject-Util-Replication-Master-v0.01/
```
patricklauer: This won't work in general since WORKDIR and other variables can be changed away from defaults.

`S=${WORKDIR}/${PN}-v${PV}` might work in all situations

einhverfr: This also breaks if I rename the pgtelemetry build.  What should I use instead of ${PV} there?

einhverfr: Never mind.  I misread the error message.

PR #87, opened 2017-08-31
Pgtelemetry upgrade

einhverfr:
allows upgrading to pgtelemetry 1.0, which has been tested more extensively on replicas and allows finer reporting of light-weight locking conditions.


patricklauer:
lgtm

PR #88, opened 2017-09-07
Perl maxmind reader

einhverfr:
This adds MaxMind::DB::Reader and all dependencies from CPAN.


einhverfr:
To test, all perl modules are dependencies of MaxMind::DB::Reader and MaxMind::DB::Reader::XS

PR #94, opened 2017-09-27
Initial commit of roleman ebuild

einhverfr:
This provides functions which make it a lot easier to manage roles in PostgreSQL (reset and add roles, create roles, manage passwords and drop roles are all currently supported).


alip:
looks good to me, admittedly nitpicking but `inherit`ing `eutils` is not necessary :)

einhverfr:
BTW, there is an odd incompatibility with the tests in 9.4 with this module so restricting tests should be done until we find a solution.  (The problem is an error message format change, not a behaviour change that affects the extension).

patricklauer:
There was a leftover Makefile modification in src_prepare, I cleaned that up and squashed the commits together.

PR #97, opened 2017-10-20
Bumping roleman version for GUC/role management

einhverfr:
Initial application is to set statement_timeout for some roles.  Could also
be used to set slow query logging or other parameters for some of our roles
as well.


alip:
lgtm.

alip:
Merged.

PR #112, opened 2017-11-17
dev-db/pg-ajbool: Rename package

patricklauer:
To avoid cluttering the namespace I suggest removing all extensions from `foo` to `pg-foo`.
This is a simple proof of concept with ajbool


einhverfr:
looks good to me.

monsieurp:
This PR is 4 year old. Is it still relevant in 2021?

monsieurp:
Got a confirmation by @patricklauer on Slack that the PR is old and can be closed. We can revisit it later if needed.

PR #117, opened 2018-01-12
Add rex-1.6.0.1

alip:


app-admin/rex/rex-1.6.0.1.ebuild
```diff
@@ -0,0 +1,145 @@
...
+	GITHUB_REPO=Rex
+	EGIT_REPO_URI="https://github.com/${GITHUB_USER}/${GITHUB_REPO}.git"
+	EGIT_BRANCH="development-1.x"
+	EGIT_CHECKOUT_DIR="${WORKDIR}/${PN}-git"
+	VCS_ECLASS="git-r3"
+else
+	# This is intentional to stop perl-module.eclass doing magic things when it
+	# shouldn't. Like making ${S} contain "Rex" when the git clone has "rex"
+	# Also prevents perl-module.eclass provisioning SRC_URI
+	DIST_AUTHOR=JFRIED
```
patricklauer: That doesn't look right :)

alip: Since we override `SRC_URI` below, this is functionally void. I'd rather keep things out of CPAN till we substantially diverge from upstream merely as a subtextual sign to indicate our will to cooperate :)

PR #119, opened 2018-02-12
dev-db/metabase-bin: ebuild, init scripts, config

65278:
as requested by Wiktor, new branch with a single commit.

Package is required for this issue: https://github.com/adjust/infrastructure/issues/2595


dev-db/metabase-bin/metabase-bin-0.27.2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	eapply_user
+}
+
+src_install() {
+	keepdir /var/log/metabase
+	fowners metabase:metabase /var/log/metabase
+
+	dodir "${INSTALL_DIR}/"
+	cp -pRP * "${ED}/${INSTALL_DIR}" || die
+	fowners -R metabase:metabase "${INSTALL_DIR}"
```
alip: We could also call `fperms` to ensure sanity over permissions so everything is nice and explicit in src_install.

dev-db/metabase-bin/metabase-bin-0.27.2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+
+pkg_setup() {
+	enewgroup metabase
+	enewuser metabase -1 -1 /opt/metabase metabase
+}
+
+src_unpack() {
+	#cp because we don't actually unpack jars
+	mkdir -p "${WORKDIR}/${MY_P}/bin"
+	mkdir -p "${WORKDIR}/${MY_P}/lib"
```
alip: mkdir needs to die on failure or be replaced with an ebuild helper.

dev-db/metabase-bin/files/metabase-server-start.sh
```diff
@@ -0,0 +1,88 @@
...
+	export MB_PASSWORD_STRENGTH="${METABASE_PASSWORD_STRENGTH}"
+
+[ -n "${METABASE_PASSWORD_LENGTH}" ] && \
+	export MB_PASSWORD_LENGTH="${METABASE_PASSWORD_LENGTH}"
+
+if [ -n "${METABASE_ENCRYPTION_SECRET_KEY}" ]
+then
+	echo ${METABASE_ENCRYPTION_SECRET_KEY} | \
+		base64 --decode &> /dev/null || \
+		exit_error "${METABASE_ENCRYPTION_SECRET_KEY}"
```
alip: Why not something like `bad secret key, invalid base64` for a little context?
Avoiding to print out the secret key to standard error is another good idea, imo.

PR #126, opened 2018-03-06
dev-lang/R-3.4.1: ebuild/patch that removes cairos dependency on X

65278:
Needed R to be able to plot without X11.

Eventually this is stuff for upstreaming, but the sanity of cairo-xlib.h inclusion is not given at the moment (i.e. I don't know what happens if you build --with-x).


dev-lang/R/R-3.4.1.ebuild
```diff
@@ -0,0 +1,205 @@
...
+	# fix HTML links to manual (gentoo bug #273957)
+	sed -e 's:\.\./manual/:manual/:g' \
+		-i $(grep -Flr ../manual/ doc) || die "sed for HTML links failed"
+
+	use lapack && \
+		export LAPACK_LIBS="$($(tc-getPKG_CONFIG) --libs lapack)"
+
+	if use X; then
+		export R_BROWSER="$(type -p xdg-open)"
+		export R_PDFVIEWER="$(type -p xdg-open)"
```
alip: Nitpick(ish): `-P` is preferred over `-p` for ebuilds mostly due to historical reasons (-P dates back to bash-2.05b).

dev-lang/R/R-3.4.1.ebuild
```diff
@@ -0,0 +1,205 @@
...
+	doenvd 99R
+	newbashcomp "${DISTDIR}"/${PN}-${BCPV}.bash_completion ${PN}
+	# The buildsystem has a different understanding of install_names than
+	# we require.  Since it builds modules like shared objects (wrong), many
+	# objects (all modules) get an incorrect install_name.  Fixing the build
+	# system here is not really trivial.
+	if [[ ${CHOST} == *-darwin* ]] ; then
+		local mod
+		pushd "${ED}"/usr/$(get_libdir)/R > /dev/null
+		for mod in $(find . -name "*.dylib") ; do
```
alip: Nitpick: Hello darwin. Why not -print0?

65278:
I'll put the relevant gentoo bug report here: https://bugs.gentoo.org/638778

Also, I have commented there that we may want to simply force cairo X dependency if we set dev-lang/R X

PR #136, opened 2018-04-15
sys-kernel: Bump, now with mitigation fixes behind a useflag

patricklauer:
Package-Manager: Portage-2.3.28, Repoman-2.3.9


patricklauer:
It looks like 4.15+ break network interface bonding

https://bugs.gentoo.org/653668

PR #137, opened 2018-04-23
Version bump to ephemeralpg-2.5 (aka pg_tmp)

alip:
This fixes the problem we've been recently having on jenkins wrt. pg_tmp failing to stop postgres and making the tests hang waiting for timeout to expire.


patricklauer:
:+1:

PR #157, opened 2018-06-08
Add postgresql-10.4-r71 ebuild with -dbg SLOT suffix for cassert

alip:
The single change here apart from the other ebuilds is the SLOT name and IUSE does not have cassert anymore as econf unconditionally calls --enable-cassert.

Revision is picked specifically to be smaller than 101 to prevent any
potential problems during dependency resolution.

Admittedly, suffixing the SLOT is not the optimal way to do this. In the
future we could use virtuals for more fine-grained install control. See
the relevant issue for more context. Ref: adjust/infrastructure#3345

### DIFF
--- a/dev-db/postgresql/postgresql-10.4-r71.ebuild
+++ b/dev-db/postgresql/postgresql-10.4-r71.ebuild
@@ -1,0 +1,470 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="6"
+
+PYTHON_COMPAT=( python2_7 python3_{4,5,6} )
+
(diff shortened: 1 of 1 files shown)


alip:
> Well, technically a slot called 10-dbg is okay. A revision smaller than 101 is ok too, and postgresql handles slotting nicely.

Using different blocks of revisions for different flavours of postgresql (custom-patched, cassert enabled) has its pros and cons as well (non-intuitive, hard to maintain for high numbers of different flavours). Both, imho, are not optimal.

> On the other hand, it's a tad contrived. If it ever becomes a serious problem we can still introduce a virtual.

Agreed. Let's think about this problem when it becomes a problem.

PR #158, opened 2018-06-12
New package: Postgresql export

alip:
The init script is not tested and may need tweaking during deploy.
Following @ikitiki's advise, I decided to add a blocker on prometheus-postgres_exporter rather than creating a new user for this one. There are no collisions apart from the log directory and the user name.


roa:
lgtm. failing tests are because of other ebuilds.

ikitiki:
ping

PR #160, opened 2018-06-12
shiny-server: user shiny mustn't own shiny-server config

65278:
Shouldn't be able to change its own config (what if somebody exploits shiny server).


alip:
lgtm. Travis moans about xproto (xorg has decided to merge the split packages back again) which is unrelated.

patricklauer:
Rebase on master should make travis happy

PR #162, opened 2018-07-06
sys-fs/zfs-kmod, sys-kernel/spl: make ebuilds accept kernel 4.17

65278:
as far as I can tell, the only reason kernel is restricted to 4.16
is that zfs 0.7.9 was released before 4.17 was released

The ebuilds only differ in kernel version check.


patricklauer:
RESO UPSTREAM :vatican_city:

PR #169, opened 2018-08-09
net-analyzer/luigi: new package

wiktor-b:
wip/initial imlementation. I will likely add an openrc script.


net-analyzer/luigi/files/luigi-2.7.6.init
```diff
@@ -0,0 +1,11 @@
+#!/sbin/openrc-run
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+command="/usr/bin/luigid"
+start_stop_daemon_args="-bm"
+pidfile="/var/run/luigi.pid"
```
patricklauer: /var/run is a symlink to /run these days, so best to use /run directly (can't stop progress ;) )

patricklauer:
:+1:

net-analyzer/luigi/files/luigi-2.7.6.init
wiktor-b: nobody can make me!

PR #172, opened 2018-08-11
sys-kernel: Bump to 4.17.14

patricklauer:
Package-Manager: Portage-2.3.44, Repoman-2.3.10


patricklauer:
Now with 4.18 goodness

PR #178, opened 2018-08-22
net-analyzer/luigi: revbump

wiktor-b:
configurable logfile and config file, save state across reboots

### DIFF
--- a/net-analyzer/luigi/files/luigi-2.7.6.init
+++ b/net-analyzer/luigi/files/luigi-2.7.6.init
@@ -3,7 +3,9 @@
 # Distributed under the terms of the GNU General Public License v2
 
 command="/usr/bin/luigid"
-start_stop_daemon_args="-bm"
+log=${logfile:-"/var/log/luigi.log"}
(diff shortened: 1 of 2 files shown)


net-analyzer/luigi/files/luigi-2.7.6.init
```diff
@@ -3,7 +3,9 @@
 # Distributed under the terms of the GNU General Public License v2
 
 command="/usr/bin/luigid"
-start_stop_daemon_args="-bm"
+log=${logfile:"/var/log/luigi.log"}
+cfg=${cfgfile:"/etc/luigi/client.cfg"}
```
alip: These should be `:-` rather than `:` no?
Funny note, `bash -n` does not catch this and this fires a syntax error only if `logfile` is defined otherwise `log` is an empty string.

The quotes are not necessary but it doesn't hurt to keep them.

wiktor-b: fixed, good catch!

PR #182, opened 2018-09-06
dev-util/android-ndk: add 15c

alip:
Package-Manager: Portage-2.3.49, Repoman-2.3.10

This is identical to android-ndk-16b ebuild except that the repoman warning
below is fixed in RDPEND (slot `0/5` is no longer present in gentoo-x86)

```
  dependency.unknown            1
   dev-util/android-ndk/android-ndk-15c.ebuild: RDEPEND: sys-libs/ncurses:0/5[tinfo]
```

Ref: adjust/infrastructure#4002


patricklauer:
:+1:

PR #185, opened 2018-09-16
schaufel-0.2.ebuild: exactly what it says on the tin

65278:
This PR is urgent: Schaufel 0.1 doesn't compile with gcc 7.3 because of overzealous Werror pedantic.


patricklauer:
:+1:

PR #197, opened 2018-10-26
Bagger 0 2

65278:
This PR should allow us to provision baggers automatically, and the revision 0.2 adds amongst other things the ability to have namespaces (infixes) per schaufel/listener for multidc.

On top of that:
- 
- Oleksis fix for the insert trigger, which hopefully only warns instead of segfaulting as pfree does 
- A pull.pl limiter
- A fix for table generation/insert trigger updates already deployed in production


patricklauer:
:+1:

PR #201, opened 2018-12-10
schaufel: ebuild and manifest for 0.3

65278:
Highlights:
 - Schaufel now does binary messages.
 - kafka error messages are logged no more
 - logging no longer logs a null byte


patricklauer:
:+1:

PR #205, opened 2019-01-07
sys-cluster/spark-bin: ebuild for 2.4.0

maxeaubrey:
Same as gentoo, but 2.4.0


sys-cluster/spark-bin/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
+    <email>java@gentoo.org</email>
```
patricklauer: Maybe we should fix that to ops@adjust.com so no one mistakenly files a bug with our custom ebuild? :)

maxeaubrey: Fair point. Amended!

patricklauer:
:+1:

PR #217, opened 2019-02-28
schaufel-0.4: ebuild and Manifest

65278:
Small new release. Nothing done but a worker thread per postgres thread to commit messages periodically.


patricklauer:
:+1:

PR #220, opened 2019-03-10
dev-db/pipelinedb: Initial ebuild

patricklauer:
Package-Manager: Portage-2.3.62, Repoman-2.3.12


monsieurp:
@patricklauer same question :)

PR #222, opened 2019-03-13
Certbot NS1 plugin

maxeaubrey:
certbot-dns-nsone plugin and related dependencies.


dev-python/tldextract/tldextract-2.2.1.ebuild
```diff
@@ -0,0 +1,26 @@
...
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
+"
+RDEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
```
alip: Is it necessary to RDEPEND on setuptools? It's unlikely to be used at runtime.

maxeaubrey: Pip *claims* it needs it, and i didn't spend any time verifying - i doubt it too, but :woman_shrugging:.

alip: ah well, let's leave it to upstream then.

PR #223, opened 2019-03-14
net-analyzer/logdna-agent: ebuild and manifest for 1.5.6

maxeaubrey:


net-analyzer/logdna-agent/files/logdna-agent
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+  
+cd $LOGDNA_HOME
```
alip: I think it's a good idea to `|| exit 1` or do `set -e` so the script won't execute further in case of chdir()'ing into the directory fails.

net-analyzer/logdna-agent/files/logdna-agent
```diff
@@ -0,0 +1,4 @@
...
+node index.js "$@"
```
alip: You could prepend this with `exec` so we don't keep the shell process running.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+SRC_URI="https://github.com/logdna/logdna-agent/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="net-libs/nodejs"
+RDEPEND="${DEPEND}"
+BDEPEND=""
```
alip: This line should not be necessary.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+
+DOCS=( license.txt README.md )
+
+src_compile() {
+	npm install
```
alip: `|| die "npm install failed`"

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	cp scripts/openrc-init scripts/logdna-agent
```
alip: `|| die "cp failed"` and it's good practise to use cp with `-p`

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="net-libs/nodejs"
+RDEPEND="${DEPEND}"
+BDEPEND=""
+
+DOCS=( license.txt README.md )
```
alip: It's unnecessary to install `license.txt` under `/usr/share/doc` the `LICENSE` variable handles the issue by directing the user to read the relevant license under `/usr/portage/licenses`.

net-analyzer/logdna-agent/files/logdna-agent
65278: Shouldn't we quote this here and fail if `cd` fails?

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
65278: Is an empty bdepend required?

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
65278: Paranoid, but we should error out if either of these fail.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+DESCRIPTION="LogDNA's collector agent which streams log files to your LogDNA account."
+HOMEPAGE="https://github.com/logdna/logdna-agent"
+SRC_URI="https://github.com/logdna/logdna-agent/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="net-libs/nodejs"
```
65278: Do we have a minimum version of node this requires? node is in quite a state of flux.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+src_compile() {
+	npm install
+	cp scripts/openrc-init scripts/logdna-agent
+}
+
+src_install() {
+	dodir logdna-agent
+	into usr/lib/logdna-agent
+
+	insinto usr/lib/logdna-agent
```
alip: It's a good idea to use `$(get_libdir)` rather than merely `lib`.
See: https://devmanual.gentoo.org/archs/amd64/index.html#libdir-links
Note ^^ is a bit old, iirc starting with EAPI 6 you don't need to inherit multilib eclass to get `get_libdir` function.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	doins -r node_modules
+	doins index.js
+	doins package.json
+
+	doinitd scripts/logdna-agent
+
+	dobin "${FILESDIR}"/logdna-agent
+	doenvd "${FILESDIR}"/99logdna-agent
+
+	dosym "${ED%/}"/usr/lib/logdna-agent/bin/logdna-agent /usr/bin/logdna-agent
```
alip: ditto about `/usr/lib` -> `/usr/$(get_libdir)`

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
maxeaubrey: We do not, hence the generic dependency.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
maxeaubrey: No, oversight on my part.

maxeaubrey:
Thank you for the notes, will address.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+DOCS=( README.md )
+
+src_compile() {
+	npm install || die "npm install failed"
+	cp -p scripts/openrc-init scripts/logdna-agent || die "cp failed"
+}
+
+src_install() {
+	dodir logdna-agent
+	into usr/$(get_libdir)/logdna-agent
```
alip: typo? `into` is not defined for me.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	insinto usr/$(get_libdir)/logdna-agent
+	doins -r lib
+	doins -r node_modules
+	doins index.js
+	doins package.json
+
+	doinitd scripts/logdna-agent
+
+	dobin "${FILESDIR}"/logdna-agent
+	sed -i "s/@LIBDIR@/$(get_libdir)/" "${FILESDIR}"/99logdna-agent
```
alip: `|| die sed failed`
Note, all `do*` functions die on their own on failure, everything else needs a ` || die`.

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
maxeaubrey: No, i went by this, first row: https://devmanual.gentoo.org/function-reference/install-functions/index.html

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
maxeaubrey: Ah, right, will take care to always have those, thanks!

net-analyzer/logdna-agent/logdna-agent-1.5.6.ebuild
alip: til, thanks for the reference.

PR #229, opened 2019-04-03
sys-cluster/nomad: 0.9.0-rc1 with patches

maxeaubrey:
Newer version of Nomad. Currently a release candidate, should be replaceable with stable ~soon.

Patches in support for mounting volumes from the system into isolated drivers, sourced from this upstream branch: https://github.com/hashicorp/nomad/compare/dani/host-volumes

Tested as principially working.

cc: @roa for written confirmation that the use of an outside patch is OK in this case.


roa:
Ack

patricklauer:
:+1:

PR #233, opened 2019-04-10
Revert "app-metrics/collectd-haproxy: HAProxy plugin for Collectd"

matt604-adjust:
Reverts adjust/gentoo-overlay#230

m!


matt604-adjust:
Argh. This is not what I was expecting. Derp.

m!

PR #238, opened 2019-04-25
Metabase clickhouse plugin

maxeaubrey:
For adjust/infrastructure#5252.


dev-db/metabase-clickhouse-driver-bin/metabase-clickhouse-driver-bin-0.1.ebuild
```diff
@@ -0,0 +1,37 @@
+# Copyright 1999-2019 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+inherit user
```
alip: This eclass is not used in the ebuild afaict.

dev-db/metabase-bin/files/metabase-server-start.sh
```diff
@@ -82,6 +82,11 @@ then
 	export MB_JETTY_SSL_Keystore_Password="${METABASE_SSL_KEYSTORE_PASSWORD}"
 fi
 
+if [ -n "${METABASE_PLUGINS_DIR}" ] && [ -d "${METABASE_PLUGINS_DIR}" ]
```
alip: `-d` alone should be enough, no?

dev-db/metabase-clickhouse-driver-bin/metabase-clickhouse-driver-bin-0.1.ebuild
```diff
@@ -0,0 +1,37 @@
...
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+RDEPEND="
+	>=dev-db/metabase-bin-0.32.0
+"
+
+DEPEND="${RDEPEND}"
```
alip: This is also unnecessary (we merely cp during build time so don't necessarily need metabase-bin present).

dev-db/metabase-clickhouse-driver-bin/metabase-clickhouse-driver-bin-0.1.ebuild
maxeaubrey: You're right, will amend.

dev-db/metabase-clickhouse-driver-bin/metabase-clickhouse-driver-bin-0.1.ebuild
maxeaubrey: True, but i thought it would be nice to keep the logical order of packages.

dev-db/metabase-bin/files/metabase-server-start.sh
maxeaubrey: Ah, yup. No need to double check if empty

PR #239, opened 2019-04-26
dev-db/metabase* fixup

maxeaubrey:
Turns out the docs lie, but plugins are autoloaded from lib/plugins anyway.


patricklauer:
:+1:

PR #240, opened 2019-04-29
dev-qt/qtwebkit: Qtwebkit removing gstreamer

matt604-adjust:
Removing gstreamer as current version is masked and we don't seem to be using it.

### DIFF
--- a/dev-qt/qtwebkit/qtwebkit-4.8.5-r100.ebuild
+++ b/dev-qt/qtwebkit/qtwebkit-4.8.5-r100.ebuild
@@ -14,7 +14,7 @@ if [[ ${QT4_BUILD_TYPE} == release ]]; then
 	KEYWORDS="~amd64"
 fi
 
-IUSE="+gstreamer icu +jit wkhtmltopdf"
+IUSE="icu +jit wkhtmltopdf"
 
 # libxml2[!icu?] is needed for bugs 407315 and 411091
 DEPEND="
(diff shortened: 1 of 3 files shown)


dev-qt/qtwebkit/qtwebkit-4.8.5-r100.ebuild
```diff
@@ -75,7 +69,6 @@ multilib_src_configure() {
 		-system-sqlite
 		$(qt_use icu)
 		$(qt_use jit javascript-jit)
-		$(use gstreamer || echo -DENABLE_VIDEO=0)
```
maxeaubrey: i think the 2nd part of this needs to stay.

dev-qt/qtwebkit/qtwebkit-4.8.7-r100.ebuild
```diff
@@ -73,7 +67,6 @@ multilib_src_configure() {
 		-system-sqlite
 		$(qt_use icu)
 		$(qt_use jit javascript-jit)
-		$(use gstreamer || echo -DENABLE_VIDEO=0)
```
maxeaubrey: Ditto

matt604-adjust: Derp. I misread the "or". Updated.

dev-qt/qtwebkit/qtwebkit-4.8.5-r100.ebuild
matt604-adjust: Derp. I misread the "or". Updated.

PR #241, opened 2019-04-29
app-metrics/collectd-haproxy: Clean up broken path. Use defaults.

matt604-adjust:


maxeaubrey:
But strictly speaking, at that point you can drop the function redefinition

PR #247, opened 2019-05-11
Postgresql 10.8

patricklauer:


dev-db/bagger-data/bagger-data-0.1.ebuild
```diff
@@ -1,8 +1,8 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2019 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=5
-POSTGRES_COMPAT=( "9.6" "10" )
```
65278: We require postgres 9.6 compat for bagger-{1..30}

dev-db/postgresql/postgresql-9.6.9-r101.ebuild
```diff
@@ -1,486 +0,0 @@
-# Copyright 1999-2018 Gentoo Foundation
```
65278: We require a postgres 9.6 ebuild because we patch gin to be able to have > 32 indices.

patricklauer:
Left pg-9.6 ebuilds alone, should be fine

PR #250, opened 2019-05-15
Schaufel 0 6

65278:
Newest schaufel. Should be backwards compatible with older schaufels.
Removed ebuilds for memory leaking versions while I was at it.


patricklauer:
:+1:

PR #254, opened 2019-05-20
An ebuild for libarrow/libparquet

sumerman:
An ebuild for libarrow/libparquet that are required for Postgres FDW (ebuild is coming soon).


dev-libs/libarrow/libarrow-0.13.0.ebuild
```diff
@@ -0,0 +1,40 @@
+EAPI=7
```
maxeaubrey: Needs a copyright header.

dev-libs/libarrow/libarrow-0.13.0.ebuild
```diff
@@ -0,0 +1,40 @@
...
+inherit eutils cmake-utils 
+
+ARROW_VSN="0.13.0"
```
maxeaubrey: Why use this instead of `${PV}`?

sumerman: I didn't know about it at first and then forgot to change it. Good catch! Fixed.

dev-libs/libarrow/libarrow-0.13.0.ebuild
sumerman: done

maxeaubrey:
The main reason it's failing right now is the +x on the ebuild. If you want to fix up the warnings, there's some whitespace and an extraneous newline at the end too :)

Besides that, i cannot get it to build:

```
CMake Error at thrift_ep-stamp/download-thrift_ep.cmake:157 (message):
  Each download failed!

    error: downloading 'http://archive.apache.org/dist/thrift/0.12.0/thrift-0.12.0.tar.gz' failed
         status_code: 6
         status_string: "Couldn't resolve host name"
         log:
         --- LOG BEGIN ---
         getaddrinfo(3) failed for archive.apache.org:80
```

Downloads cannot happen anywhere outside of src_unpack afaik, but why is emake handling this at all?

sumerman:
Thank you! The error message was quite cryptic enough for me to miss it. It works on inside VM on my machine, but you are right — all those deps should be pulled by explicitly. `emake` does it because it calls into `cmake` and `libarrow` has some magic dependency handling.

maxeaubrey:
Closing in favour of #268.

PR #258, opened 2019-05-22
sys-cluster/hive-standalone-metastore-bin: ebuild and manifest for 3.0.0

maxeaubrey:
Depends on #257. Passes when merged on top of that since then the dep is there.

Many thanks to @alip for help with figuring out my problems with `JAVA_HOME` in the init script.


sys-cluster/hive-standalone-metastore-bin/hive-standalone-metastore-bin-3.0.0.ebuild
```diff
@@ -0,0 +1,70 @@
...
+SRC_URI="https://www-eu.apache.org/dist/hive/${MY_PN}-${PV}/${MY_P}.tar.gz http://central.maven.org/maven2/org/apache/logging/log4j/log4j-web/2.11.2/log4j-web-2.11.2.jar"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
+RDEPEND="
+	>=virtual/jre-1.8
+	~sys-cluster/hadoop-bin-3.2.0
```
alip: nitpick: could move depend up and use ${DEPEND} rather than repeating hadoop (is common practise).

PR #259, opened 2019-05-23
dev-db/clickhouse: fix log dir permissions in ebuild

maxeaubrey:


dev-db/clickhouse/clickhouse-19.5.3.8.ebuild
```diff
@@ -192,10 +192,10 @@ src_install() {
 	fi
 
 	keepdir /var/log/clickhouse-server
-	chown clickhouse:clickhouse "${D}"/var/log/clickhouse-server
+	fowners -R clickhouse:clickhouse /var/log/clickhouse-server
```
alip: Judging by `pkg_setup` you want to wrap this into a `use server` conditional as well else the clickhouse user won't exist on the system so `fowners` will fail.

maxeaubrey: Thanks, good catch! `server` might be default but it still could be turned off. Will fixup on merge.

alip: Indeed, sgtm.

PR #261, opened 2019-05-24
jupyterhub and dependencies ebuild

maxeaubrey:
This still has some issues with spawning notebooks, but that doesn't look related to packaging issues.

Doesn't pass repoman so not up for review yet.


maxeaubrey:
No progress for a while, closing.

PR #264, opened 2019-05-29
net-analyzer/zabbix: version bump to 4.0.7

alip:
Package-Manager: Portage-2.3.62, Repoman-2.3.11

Latest ::gentoo is 4.0.6,
6..7 has the fix below which is enough reason to be paranoid:

> ...G...PS. [ZBX-15955] fixed security vulnerability - accepting connections from not allowed addresses (Andris, vso)


patricklauer:
Upstream

PR #265, opened 2019-06-04
Trex compat

65278:
Added the standardized offensive useflag to restore compatibility with old deployments


65278:
This PR also reduces program name typing time by at least 33%.

maxeaubrey:
i don't see why this is at all needed in the ebuild. If your personal use requires this, you can achieve the same using an alias.

PR #266, opened 2019-06-04
sys-cluster/nomad: remove old versions, fixed init script to somethin…

maxeaubrey:
…g more vaguely correct.

A lot of these changes should also be upstreamed for the consul init script.


sys-cluster/nomad/files/nomad.initd
```diff
@@ -22,6 +19,6 @@ depend() {
 
 reload() {
 	ebegin "Reloading ${RC_SVCNAME}"
-	${command} reload
+	start-stop-daemon --signal HUP --pidfile
```
alip: `--pidfile` requires an argument.

maxeaubrey: Ta, thanks for spotting. Will fix.

PR #267, opened 2019-06-04
Schaufel logrotate

65278:
We might find this addition useful. The su thing will work for now, I think. It may become a problem when we let schaufel write to /dev/log

### DIFF
--- a/app-admin/schaufel/schaufel-0.6.ebuild
+++ b/app-admin/schaufel/schaufel-0.6.ebuild
@@ -55,6 +55,9 @@ src_install() {
 	export DOCDIR="${PREFIX}/share/doc/${P}"
 	default
 
+	insinto /etc/logrotate.d
+	newins "${FILESDIR}/${PN}.logrotate" ${PN}
+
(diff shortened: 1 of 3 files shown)


app-admin/schaufel/schaufel-0.6.ebuild
```diff
@@ -55,6 +55,9 @@ src_install() {
 	export DOCDIR="${PREFIX}/share/doc/${P}"
 	default
 
+	insinto /etc/logrotate.d
```
alip: It's common practise to install logrotate scripts conditionally with `use logrotate`.

65278: It'd be good practice. Common it is not. euse -i logrotate comes up with 3 packages.
That is not to say I won't indulge you. Should I set logrotate as standard enabled or change our default make.conf useflags?

alip: TIL, my knowledge seems to be outdated. Looking at the tree most ebuilds do this unconditionally these days. Let's just follow the trend then.

PR #268, opened 2019-06-07
Add dev-libs/apache-arrow

zilder:
Adds `libarrow` and `libparquet` needed by `parquet_fdw` extension.

ps: i realized too late that there had already been PR #254


zilder:
I will probably need to add a package for apache-thrift to this repo too because arrow depends on it and i can't find it in gentoo repo

zilder:
Also arrow depends on a specific version of jemalloc. It doesn't check for existing library in the system and download it anyway:
```
  # We only use a vendored jemalloc as we want to control its version.
  # Also our build of jemalloc is specially prefixed so that it will not
  # conflict with the default allocator as well as other jemalloc
  # installations.
```
Should i add a package for this specific version of jemalloc and add it as a dependency? In this case i will also need to create a patch for arrow's cmakefile to make it use system library.

maxeaubrey:
You can configure the ebuild to require a specific vesion.

> Also our build of jemalloc is specially prefixed so that it will not conflict with the default allocator as well as other jemalloc installations.

Sounds to me like you can package it as `dev-libs/arrow-jemalloc` (or whatever they're calling it), or as jemalloc in a different slot (but i don't think it's an option without also changing standard jemalloc builds to use slots?) - someone else can weigh in on what's best.

zilder:
Added dev-libs/thrift package that is required by Arrow.
Added dependencies list to dev-libs/apache-arrow.

Also Arrow requires a specific version of jemalloc and wants to build it itself with a specific prefix (it doesn't have an option to use system library). So i added a rule in ebuild that downloads jemalloc distribution and hands it to arrow build system.

dev-libs/thrift/thrift-0.12.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+DESCRIPTION="Apache Thrift"
+HOMEPAGE="https://thrift.apache.org/"
+SRC_URI="https://archive.apache.org/dist/thrift/${PV}/${P}.tar.gz"
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND=""
+RDEPEND="${DEPEND}"
+BDEPEND=""
```
maxeaubrey: See boost comment.

zilder: > Any reason you're now pulling from git, and building HEAD?

Yes, latest release doesn't allow to specify arbitrary path to jemalloc source. It is quite recent change, not yet released. After release i'll add separate ebuild.

> FYI you can pull a specific version of apache-arrow and jemalloc alongside it, in a single SRC_URI, which i feel would be the better way to do this.

As I said arrow doesn't have option to be built with system jemalloc. Probably it's possible to patch arrow to be able to do so. But i don't know whether it worth it.

> Note on naming: any reason to go with apache-arrow over libarrow like in the other PR?

This package builds two libs: libarrow and libparquet. I think it would be misleading to name it libarrow.

> This agrees you need boost: https://thrift.apache.org/docs/install/

Thanks! It seems that it also requires bison and flex. Will add to dependencies.

zilder: i think it worth mentioning that arrow doesn't install jemalloc anywhere in the system. I guess it uses static linking

maxeaubrey: Thanks for the explanation! In that case, maybe `EGIT_COMMIT`? But hopefully a new release will come soon and we can pin it to that, so it might not be worth the bother.

As far as `flex` and `bison` go, i didn't provide them and it still built fine, but definitely please verify that and the `libssl-dev` mention on that page. Should be fine after that.

zilder: done!

65278:
Please include my addons to the thrift ebuild or wait for me to add them.

lemanchester:
I have added but I guess it was forced on the top! :joy:

65278:
Okay, I think thrift is in a good state now. Won't look at arrow, can merge.

PR #271, opened 2019-06-14
bagger-tools init: set producer jobs when using a config

65278:
Found the one bug in my deployment.
This will set producer threads as intended when using a config. And this only happened because the naming convention doesn't make it clear which value applies to consumers and which to producers.


patricklauer:
:+1:

PR #272, opened 2019-06-20
dev-db/postgresql: Bump

patricklauer:
Package-Manager: Portage-2.3.67, Repoman-2.3.14


dev-db/postgresql/postgresql-10.9.ebuild
```diff
@@ -0,0 +1,469 @@
+# Copyright 1999-2019 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="6"
+
+PYTHON_COMPAT=( python2_7 python3_{4,5,6} )
```
maxeaubrey: Maybe could drop 3.4? :)

PR #277, opened 2019-07-02
sys-cluster/ceph: pull from gentoo to fix ebuild and init

maxeaubrey:
Pulled from gentoo to resolve while we wait https://bugs.gentoo.org/688828


maxeaubrey:
Changes are line 251 of ebuild (added force) and r12 init script.

PR #309, opened 2019-09-25
dev-libs/librdkafka: lz4-compat useflag for the old kafka cluster

65278:
At least some api_version 2 kafka servers support lz4 (ours does).
```>dev-libs/librdkafka-1.0.0-rc1```  binds lz4 compatibility to api_version 3. Therefore schaufel fails when talking to the old cluster.
Since I'm quite certain this patch is non upstreamable, I made an ebuild. By the request of wiktor, I made it a useflag.

Ref: https://github.com/adjust/infrastructure/issues/6254

### CODE REVIEW

File: dev-libs/librdkafka/metadata.xml [hunk 1]
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>whissi@gentoo.org</email>
```
maxeaubrey: This should be maybe changed tho :)

PR #314, opened 2019-10-13
Fixups

patricklauer:


maxeaubrey:
Um, `prometheus-node_exporter` is in use.

PR #319, opened 2019-10-31
app-misc/go-carbon: initial ebuild for 0.14.0

maxeaubrey:
Ebuild for https://github.com/lomik/go-carbon

Default `go-carbon.conf` comes from running `go-carbon -config-print-default`
Default `storage-schemas.conf` comes from carbon-cache examples https://github.com/graphite-project/carbon/blob/master/conf/storage-schemas.conf.example


maxeaubrey:
Actually, one thing i did change in `go-carbon.conf` from the default is logging to stderr for the sake of having it go thru openrc since that's doing logging anyway.

app-misc/go-carbon/files/go-carbon.conf
```diff
@@ -0,0 +1,96 @@
+[common]
+user = "carbon"
+graph-prefix = "carbon.agents.{host}"
+metric-endpoint = "local"
+max-cpu = 1
+metric-interval = "1m0s"
+
+[whisper]
+data-dir = "/var/lib/graphite/whisper/"
```
artemyevav: It is not created by default so we have lot of `mkdir /var/lib/graphite: permission denied` errors.

maxeaubrey: `enewuser` is missing homedir definition, will fix up. Good catch!

app-misc/go-carbon/go-carbon-0.14.0.ebuild
```diff
@@ -23,7 +23,7 @@ CARBON_USER="carbon"
 
 pkg_setup() {
 	enewgroup "${CARBON_USER}"
-	enewuser "${CARBON_USER}" -1 -1 -1 "${CARBON_USER}"
+	enewuser "${CARBON_USER}" -1 -1 "/var/lib/${CARBON_USER}" "${CARBON_USER}"
```
artemyevav: `/var/lib/__graphite__`

artemyevav: All the rest is working. So pls fix the issue above and move forward.

PR #329, opened 2019-12-11
Bump: chromedriver

65278:
Chromedriver needs to be in sync with google-chrome, version wise.
Since we want to update to 78.0.3904 in https://github.com/adjust/infrastructure/issues/7048, this package needs a version bump.


65278:
Actually, gentoo has an ebuild now.

PR #349, opened 2020-02-26
sys-cluster/ceph: Updated ceph 13 ebuild, removes spdk dependency

peterbangert:
When trying to install ceph on new machine, compilation kept Werroring out due to some dependency on boost (im hazy on details) .

Patrick yak shaved it into working by removing spdk dependency

Posting this here for consistency. Discussion should be whether to replace r4 with this, or have second profile available.


patricklauer:
Since we don't use any of the spdk/dpdk bits - disabling them doesn't have a visible effect, except that it allows compilation with newer compilers.
(iow: should replace -r4 where needed)

PR #351, opened 2020-02-27
Bagger exports

65278:
Ebuild for bagger-exports, required for infrastructure #6665. Much the same as bagger-data is.


patricklauer:
```
IUSE.invalid [fatal]          1
   dev-db/bagger-exports/bagger-exports-0.0.7.ebuild: schaufel
```
I think you want to add schaufel useflag to metadata.xml to make repoman happy

PR #360, opened 2020-03-27
app-admin/consul-bin: init at 1.7.2

maxeaubrey:


app-admin/consul-bin/consul-bin-1.7.2.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+DESCRIPTION="Cluster manager from Hashicorp"
+HOMEPAGE="https://www.consul.io"
+SRC_URI="https://releases.hashicorp.com/${MY_PN}/${PV}/${MY_PN}_${PV}_linux_amd64.zip -> ${P}.zip"
+
+LICENSE="MPL-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND=""
```
65278: I'd wager this depends on glibc.

app-admin/consul-bin/consul-bin-1.7.2.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2018 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit eutils user
```
65278: Can't quite fathom what we need eutils for here.

app-admin/consul-bin/consul-bin-1.7.2.ebuild
65278: Disregard that, it's statically linked.Wow.

app-admin/consul-bin/consul-bin-1.7.2.ebuild
maxeaubrey: Correct, but i don't think this really hurts? Can rebase and fix it here and for nomad-bin too, if you'd like.

PR #369, opened 2020-05-12
Subscription service

65278:
Dependencies of the subscription service.

I skipped many features/tests to finish this. Licenses should be correct, all packages install.
Flask-admin doesn't officially support python3.6, but if it's good enough for gentoo (why did it need deprecation?) it's good enough for us.


peterbangert:
Ive installed and imported most of these in esh-sub-1,2 so i can verify it all works as intended.

PR #488, opened 2020-10-17
Gostatsd-bin: A reimplementation of statsd

65278:
Gostatsd was too hard to package (its go deptree is half of github), and a package which needs network-sandbox disabled isn't a package I particularly fancy.

Therefore, I compiled gostatsd-bin and we're hosting that now (that should be alright with its MIT license).

I've also added acct-{user,group}/gostatsd to appease the Gorny.


monsieurp:
We will not use it in the end (see infra issue) so I'm closing this PR.

PR #494, opened 2020-11-03
dev-java/icedtea: update to latest version.

monsieurp:
Hi

This PR removes the broken icedtea version and replaces it with the latest one.

Please review and merge accordingly.


monsieurp:
OK I will file another PR cause the problem runs much deeper and isn't strictly related to icedtea. A bunch of qt libs are outdated and need wiping.

PR #499, opened 2020-11-15
bagger-tools-9999: replication switch

65278:
Pre release feature: LISTENER_REPLICATION allows to toggle replication.


einhverfr (approved):
looks ok to me.  Let's test on one host first before roling it out everywhere though.

PR #502, opened 2020-12-10
Aerospike init script: Changing user/group to root

einhverfr:
Avoids manual editing.


monsieurp (approved):

patricklauer (changes requested):
That doesn't do the right thing.
Proper fix is adding these variables to conf.d where needed.

PR #503, opened 2020-12-15
sys-cluster/spark-bin: version bump.

monsieurp:
This commit:
* adds conf.d/init.d files for Spark History server.
* adds conf.d/init.d files for Spark MCD: Mesos Cluster Dispatcher service.
* sets up a SLOT for Spark.
* creates the "warden" user and required directories.
* grants the "warden" user permission on said directories.

Note: this ebuild is a copy of the same ebuild that exists in the `::gentoo` repository with some adjust-specific tweaks. We will have to install this one though.

We do need the "warden" user management part to keep the Mesos cluster happy when running Mesos jobs and avoid user permission errors.

I ran repoman with "--force" to mark this package stable for amd64 right away.


monsieurp:
This ebuild has been used to install Chronos on `lax-ab-worker-3` and works fine.

monsieurp:
This ebuild has been used to install Spark on `lax-ab-worker-{1,2,3}` and works fine.

monsieurp:
Closing in favour of #512.

PR #507, opened 2021-01-06
Change running user/group to root

blueSherii:
Issue related https://github.com/adjust/infrastructure/issues/10890

### DIFF
--- a/dev-db/aerospike-server-enterprise/aerospike-server-enterprise-4.5.0.12.ebuild
+++ b/dev-db/aerospike-server-enterprise/aerospike-server-enterprise-4.5.0.12.ebuild
@@ -65,6 +65,8 @@ src_install() {
 		doins "${FILESDIR}/${conf}"
 	done
 
+	newconfd "${FILESDIR}/aerospike" aerospike
+
 	insinto /usr/bin
 	doins usr/bin/*
 	fperms +x -R /usr/bin/asd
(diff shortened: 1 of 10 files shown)


dev-db/aerospike-server-enterprise/files/aerospike.init3
```diff
@@ -0,0 +1,51 @@
+#!/sbin/openrc-run
+
```
blueSherii, line 2: user and group removed from this section .
A new init file created to preserve the old ones

einhverfr: not sure why this is needed.  Couldn't putting relevant things in conf.d by our automation do it better?

einhverfr (changes requested):
So let's sync up on this.

My thinking is the conf.d system is probably a better place for this rather than the gentoo overlay files directly and then we can create those with our automation.  Let's discuss soon

dev-db/aerospike-server-enterprise/aerospike-server-enterprise-4.5.0.12.ebuild
```diff
@@ -65,14 +65,17 @@ src_install() {
 		doins "${FILESDIR}/${conf}"
 	done
 
+	insinto /etc/conf.d
```
monsieurp, line 68: Please use the `newconfd` function instead:
> newconfd | Install an /etc/conf.d entry using the second argument as the name

See: https://devmanual.gentoo.org/function-reference/install-functions/

monsieurp (changes requested):

blueSherii:
@einhverfr I think putting this in gentoo-overlay is better when the conf.d file is mostly static and not "server/flavour/other config" dependent

mcharriere:
from my point of view, there is no need for this PR. The issue can be solved adding a conf.d in the infra repo. If fact since this changes the default values of user/root it will also change where root is not needed (like in aero-tracker).

PR #509, opened 2021-01-15
sys-cluster/mesos: new ebuild.

monsieurp:
Apache Mesos is an open-source project to manage computer clusters.


sys-cluster/mesos/files/mesos-agent.confd
```diff
@@ -0,0 +1,32 @@
+# Apache Mesos agent node conf.d file.
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# ZooKeeper server IP.
+ZOOKEEPER_HOST="foobar.adjust.com"
```
patricklauer, line 6: Please don't hardcode real 'fake' domains in config files

sys-cluster/mesos/files/mesos-agent.confd
```diff
@@ -0,0 +1,32 @@
...
+
+# ZooKeeper TCP port.
+# Default port is 2181.
+ZOOKEEPER_PORT="2181"
+
+# Default working directory.
+MESOS_AGENT_WORK_DIR="/var/lib/warden/mesos"
```
patricklauer, line 13: why 'warden' ?

sys-cluster/mesos/files/mesos-master.confd
```diff
@@ -0,0 +1,45 @@
+# Apache Mesos master node conf.d file.
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# Mesos master node IP.
+# 0.0.0.0 means Mesos will listen on all interfaces by default.
+MESOS_MASTER_IP="0.0.0.0"
+
+# ZooKeeper server IP.
+ZOOKEEPER_HOST="foobar.adjust.com"
```
patricklauer, line 10: same as above, at least use 'example.com' or something

sys-cluster/mesos/files/mesos-master.confd
```diff
@@ -0,0 +1,45 @@
...
+
+# ZooKeeper TCP port.
+# Default port is 2181.
+ZOOKEEPER_PORT="2181"
+
+# Default cluster name.
+MESOS_MASTER_CLUSTER_NAME="'adjust Apache Mesos cluster'"
```
patricklauer, line 17: don't hardcode implementation details into generic config files please

sys-cluster/mesos/mesos-1.10.0.ebuild
```diff
@@ -0,0 +1,103 @@
...
+}
+
+src_compile() {
+	# just so that these variables are set up throughout the whole process.
+	export CC="$(tc-getCC)" CXX="$(tc-getCXX)"
+
+	# make will fail on:
+	# src/core/lib/gpr/log_linux.cc:42:13: error: ambiguating new declaration
+	# of ‘long int gettid()’
+	make -j4
```
patricklauer, line 68: hardcoding MAKEOPTS doesn't sound like a good choice

sys-cluster/mesos/mesos-1.10.0.ebuild
```diff
@@ -0,0 +1,103 @@
...
+RDEPEND="
+	>=virtual/jre-1.8"
+
+DEPEND="
+	${RDEPEND}
+	>=virtual/jdk-1.8
+	dev-java/maven-bin:3.6
+	dev-java/openjdk-bin
+	dev-libs/cyrus-sasl
+	dev-vcs/subversion
```
patricklauer, line 28: that's unexpected :)

sys-cluster/mesos/files/mesos-agent.confd
monsieurp: Fixed!

sys-cluster/mesos/files/mesos-agent.confd
monsieurp: This user account is already set up across the AB servers and used for various services such as running Spark jobs, mounting Ceph shares, etc. I have also written a small explanation about using the same account is required: https://github.com/adjust/gentoo-overlay/pull/510/commits/e2edfa13aa0310ed4afdbf65300566bda88d3f83#diff-d728260916cddcfa01b360042cb57ee84b1229a195dcd6b81e7fda5e1dc3d808R26.

sys-cluster/mesos/files/mesos-master.confd
monsieurp: Fixed.

sys-cluster/mesos/mesos-1.10.0.ebuild
monsieurp: As we already discussed on Slack, the Mesos build process is somewhat broken. Some source files are generated during the configure phase and there is nothing I can do about it unfortunately. Running `make `here does not trigger the actual build process but only starts it to get it to fail. The broken files are then patched to allow the build to be resumed later on with `emake` (see a few lines below). I know it looks funky but this workaround kind of works.

sys-cluster/mesos/mesos-1.10.0.ebuild
monsieurp: Quote from https://mesos.apache.org/documentation/latest/building/#centos-7-1:
> 'Mesos > 0.21.0' requires 'subversion > 1.8' devel package

I also wondered why but realised life's too short. You might not want to dwell too long on this question though! ;)

sys-cluster/mesos/files/mesos-master.confd
monsieurp: Fixed! Thanks.

PR #510, opened 2021-01-15
sys-cluster/chronos-bin: new ebuild.

monsieurp:
Chronos is a fault tolerant job scheduler for Mesos.


sys-cluster/chronos-bin/files/chronos.confd
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# JVM options.
+JVM_OPTS=""
+
+# ZooKeeper server IP/hostname.
+ZOOKEEPER_HOST="foobar.adjust.com"
```
patricklauer, line 8: Please change

monsieurp: Fixed.

PR #512, opened 2021-01-15
sys-cluster/spark-bin: version bump.

monsieurp:
This commit:
* adds conf.d/init.d files for Spark History server.
* adds conf.d/init.d files for Spark MCD: Mesos Cluster Dispatcher service.
* sets up a SLOT for Spark.
* creates the "warden" user and required directories.
* grants the "warden" user permission on said directories.

We do need the "warden" user management part to keep the Mesos cluster
happy when running Mesos jobs and avoid user permission errors.

I ran repoman with "--force" to mark this package stable for amd64 right away.


sys-cluster/spark-bin/files/spark-history.confd
```diff
@@ -0,0 +1,29 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+# Web UI port.
+SPARK_HISTORY_PORT="18080"
+
+# Server log.
+SPARK_HISTORY_LOG_FILE="/var/log/warden/spark-history.log"
+
+# Events directory.
+SPARK_HISTORY_EVENTS_DIRECTORY="/mnt/ceph/spark-logs"
```
patricklauer, line 12: that's a nonstandard path :)

monsieurp: Thanks! Fixed.

PR #518, opened 2021-01-21
sys-cluster/spark-bin: add more checkpath checks.

monsieurp:


sys-cluster/spark-bin/files/spark-history.confd
```diff
@@ -5,8 +5,14 @@
...
 
+# User account the Spark History process runs as.
+SPARK_HISTORY_USER="spark"
+
+# Group the Spark History process runs as.
+SPARK_HISTORY_GROUP="spark"
+
 # Server log.
-SPARK_HISTORY_LOG_FILE="/var/log/warden/spark-history.log"
+SPARK_HISTORY_LOG_FILE="/var/log/${SPARK_HISTORY_USER}/spark-history.log"
```
mcharriere, line 15: this dir is hardcoded on the ebuild. maybe you need to update it there also?

```
    for x in /var/{lib,log}/warden; do                                                                                                                                                                                
        keepdir "${x}"                                                                                                                                                                                                
        fowners warden:warden "${x}"                                                                                                                                                                                  
    done                                                                                                                                                                                                              
```

monsieurp: You are right however to the best of my knowledge there's no way to pass this value to the ebuild! :[

PR #528, opened 2021-02-03
dev-libs/apache-arrow: add cmakeargs for compression types

xmhd:
Initial fixes to apache-arrow ebuild.

Now passing some additional args to cmake for multiple compression types.

This _should_ be part of fixes for [https://github.com/adjust/infrastructure/issues/10897](url)


lemanchester (approved):
I cannot test it right now but it LGTM

xmhd:
I would be interested in implementing USE flags for gzip, zstd etc in this ebuild, and modifying the *DEPEND logic to match that.

Is there any reason we are not currently doing so?

PR #474, opened 2020-08-11
app-admin/trex: Version bump to 1.12.1

alip:
Package-Manager: Portage-3.0.1, Repoman-2.3.11

PR #534, opened 2021-02-18
app-admin/rex: Initial ebuild with adjust patches

alip:
Package-Manager: Portage-3.0.8, Repoman-2.3.11

This is rex-1.13.2 plus the adjust patches which have been included in trex-1.6.0.4 except RexOps/Rex#1471 which has been in trex master. I have carefully reviewed the changeset between current trex and latest rex and filtered out the patches which have already been fixed upstream. I have been testing this without issues for a couple of days now and there're no critical changes between the two versions so I'm relatively confident this won't kill your kittens. Still, you've been warned. Submitting this pull request with the hope to get wider testing. Note you have to uninstall trex before installing rex as they're blocking each other obviously due to file collisions.

The ebuild is identical to the one in gentoo-x86 except for the change below:

https://github.com/adjust/gentoo-overlay/blob/e0e0e9c22d06bef0cec2ea31ee640447b7425db1/app-admin/rex/rex-1.13.2.ebuild#L101-L108

All patches listed above have already been submitted upstream.
Current process of unforking can be tracked here: adjust/Rex#28.
This pull request obsoletes #474.


xmhd:
LGTM. I will test it out locally today.

alip:
After discussing with upstream we decided submitting two of the patches as modules to CPAN is a better idea.
I'll adapt this PR accordingly hence s/review/in progress/ for the time being.

alip:
Updated adjust patches and added initial ebuilds for Rex-Interface-Shell-I{drac4,lo}.

xmhd:
have been using this for a bit, and seems to be working fine.

I am merging this PR and will update the infrastructure package with || ( app-admin/rex app-admin/trex ) and we can keep trex around for bit before we decide to remove.

PR #535, opened 2021-02-24
PostgresTools-0.1: add

varilci:
Bumped PostgresTools to v0.1 for easier overlay management. Also
updated the euild to EAPI7.


dev-perl/PostgresTools/PostgresTools-0.1.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
```
xmhd, line 2: Copyright should be 1999 - $current_year Gentoo Authors

dev-perl/PostgresTools/PostgresTools-0.1.ebuild
```diff
@@ -0,0 +1,43 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
+if [[ ${PV} == 9999 ]]
+then
+    EGIT_REPO_URI="${HOMEPAGE}.git"
+else
+    SRC_URI="${HOMEPAGE}/archive/v${PVR}.tar.gz"
+fi
```
xmhd, line 19: I know it's an argument of posix vs bash, but given that we're using Gentoo and the posix compliant ebuilds argument died (sadly) with Roy Marples a decade ago, I would do like so:

```
if [[ ${PV} == 9999* ]] ; then
    EGIT_REPO_URI="https://github.com/adjust/postgres_tools.git"
else
    SRC_URI="https://github.com/adjust/postgres_tools/archive/v${PVR}.tar.gz"
fi
```

I seem to recall there being something in the Gentoo dev handbook re: [ .. ] vs [[ .. ]] as a result of Roy Marples fighting for ebuilds to be posix compliant, but I can't seem to find it right now. Every ebuild example in the handbook I can find uses the bash equivalent though fwiw.

Use of ${HOMEPAGE} and ${PN} in SRC_URI is discouraged, see [https://devmanual.gentoo.org/ebuild-writing/variables/#use-of-constant-value-variables-in-ebuilds](url)

dev-perl/PostgresTools/PostgresTools-0.1.ebuild
```diff
@@ -0,0 +1,43 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit perl-module git-r3
+
+DESCRIPTION="Adjust PostgreSQL Tools"
+HOMEPAGE="https://github.com/adjust/postgres_tools"
+SLOT="0"
+KEYWORDS="~amd64"
```
xmhd, line 11: add LICENSE="MIT" in this section

xmhd (changes requested):
```
# Copyright 2021 Gentoo Authors
```

dev-perl/PostgresTools/PostgresTools-0.1.ebuild
varilci: I was using app-vim/gentoo-syntax :sweat_smile: will fix to align with other ebuilds

dev-perl/PostgresTools/PostgresTools-0.1.ebuild
varilci: Believe it or not, I was actually looking at some other ebuilds in the repository, and writing like this seemed intuitive to me (e.g if there will be a minor change in homepage, it won't be affected too much). But since I see it's discouraged by Gentoo, I'll follow it :slightly_smiling_face:

xmhd (approved):
lgtm

PR #537, opened 2021-02-26
bagger: add postgres13 compatibility

artemyevav:
Bagger-master and bagger-data should be postgresql-13 compatible, this PR adds this feature.
Compatibility tested.

### DIFF
--- a/dev-db/bagger-master/bagger-master-0.4-r1.ebuild
+++ b/dev-db/bagger-master/bagger-master-0.4-r1.ebuild
@@ -1,0 +1,54 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+POSTGRES_COMPAT=( 9.6 10 11 12 13 )
+BUILD_DIR="${S}/sql/bagger_master"
+inherit postgres-multi
+
+DESCRIPTION="adjust bagger master extension"
+HOMEPAGE="https://github.com/adjust/bagger"
+SLOT="0"
+IUSE=""
+
+LICENSE="Unlicense"
+
+if [[ ${PV} == 9999 ]]
+then
+	inherit git-r3
+	KEYWORDS=""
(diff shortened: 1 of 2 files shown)


roa:
instead of editing the existing ebuild, can you please create an `r1` ebuild?
the docs are here: https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#file-naming-rules

That will avoid unnecessary rebuilds on systems, where the ebuild is already installed.

artemyevav:
Now it's correct?

xmhd:
```
POSTGRES_COMPAT=( "9.6" "10" "11" "12" "13" )
```

Should be

```
POSTGRES_COMPAT=( 9.6 10 11 12 13 )
```

In future could you please format your commit message as "dev-db/bagger: msg goes here" in line with the other commits + Gentoo guidelines?

I am happy to merge otherwise.

dev-db/bagger-master/bagger-master-0.4-r1.ebuild
```diff
@@ -0,0 +1,54 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+POSTGRES_COMPAT=( 9.6 10 11 13 )
```
xmhd, line 5: I missed this last time, but is 12 intentionally not added to POSTGRES_COMPAT? I can see 12 is not in that array in the bagger-master ebuilds already in the overlay. If bagger-master does indeed support 12, it would be nice to add it.

Otherwise lgtm.

artemyevav: Haven't tested it with 12

xmhd: Ok. I've been informed that if it works with 13, then it should work with 12 too. If you can add 12 to POSTGRES_COMPAT for bagger-master, then we can merge this after.

cheers

PR #540, opened 2021-03-02
dev-java/icedtea{-bin}: unfork

xmhd:
Unfork IcedTea packages from ::adjust. There are no changes from the ebuilds in ::gentoo , and we are migrating from IcedTea to OpenJDK anyway.


varilci (approved):
bombs away

PR #541, opened 2021-03-02
sys-process/vixie-cron: remove from overlay

xmhd:
Removes the following from overlay:

sys-process/vixie-cron
virtual/cron

This is in anticipation of the upcoming migration from vixie-cron to cronie.


varilci (approved):
lgtm

PR #542, opened 2021-03-03
first stab at infrastructure-meta ebuild

xmhd:
First stab at creating a meta package for all Perl modules required to use the infrastructure repository.

There are a few TODOS in there (e.g. LICENSE and HOMEPAGE), and the category + name of the package is meh, so feel free to suggest a better alternative.

If I have missed anything, please let me know and I can update  the DEPEND list.

### DIFF
--- a/virtual/adjust-infrastructure/adjust-infrastructure-0.1.ebuild
+++ b/virtual/adjust-infrastructure/adjust-infrastructure-0.1.ebuild
@@ -1,0 +1,37 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="Meta package for tools used in Adjust's infrastructure"
+
+KEYWORDS="amd64 x86"
+
+SLOT="0"
+
+IUSE=""
+
+DEPEND="
+	dev-perl/autovivification
+	dev-perl/Data-Printer
+	dev-perl/Digest-CRC
+	dev-perl/HTTP-Messages
+	dev-perl/JSON
+	dev-perl/JSON-MaybeXS
+	dev-perl/libwww-perl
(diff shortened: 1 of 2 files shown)


monsieurp:
I'd actually put the package in the `virtual` category and use a more meaningful package name, something like: `virtual/adjust-infrastructure`. What do you think?

xmhd:
> I'd actually put the package in the `virtual` category and use a more meaningful package name, something like: `virtual/adjust-infrastructure`. What do you think?

I never put it in the virtual category because afaik (from the top of my foggy head), there aren't any meta packages in there other than one or two CPAN related things for Perl. The meta packages in ::gentoo that I can remember from the top of my head are things like gnome-base/gnome and kde-plasma/plasma-meta. I am not overly bothered which category it goes in though.

wrt naming, adjust-infrastructure works for me. 

Failing checks are due to HOMEPAGE and LICENSE having 'TODO' in them. Will fix up in the morning.

alip:
+1 for virtual CATEGORY & adjust-infrastructure PN.

xmhd:
Renamed the PN and moved to virtual category. Also removed HOMEPAGE and LICENSE, so the errors on this PR will be gone now :-)

alip:
About CI checks: If we want to keep using stable keywords unconditionally, it makes sense to disable the repoman check which throws a fatal error for adding new ebuilds with stable keywords. @xmhd, do you know if there's a way to do that?

xmhd:
> About CI checks: If we want to keep using stable keywords unconditionally, it makes sense to disable the repoman check which throws a fatal error for adding new ebuilds with stable keywords. @xmhd, do you know if there's a way to do that?

There is a repoman --straight-to-stable option which afaik, 'just works' as long as all of the dependencies are also stable. So I will need to mark all of the packages in my perl pull request as stable before this one will pass QA.

PR #545, opened 2021-03-05
dev-db/clickhouse: PV bump + fix-ups

xmhd:
As per title. First pull request for fixing up Clickhouse.

- PV bumped to the latest release from 19.6 branch.
- Patches added to fix build errors when using newer system libraries.
- Cleaned old versions

There is no logic change in the ebuild other than adding eapply for patches.

I have tested locally on an ~amd64 Gentoo Linux machine.


varilci (approved):
lgtm

PR #546, opened 2021-03-10
app-admin/bagger-tools: add replication flag to the default init script

artemyevav:
`LISTENER_REPLICATION` parameter is not used by default init script. This PR adds replication control from #499 to the current version.
@65278 @roa


xmhd (approved):
lgtm

PR #549, opened 2021-03-11
import some packages removed from ::gentoo

xmhd:
Import some older packages removed from ::gentoo that we still require for some local testing.


varilci (approved):
lgtm

PR #550, opened 2021-03-17
dev-db/clickhouse: remove parallel jobs cmake args

xmhd:
remove parallel jobs cmake arg from clickhouse ebuild, it causes issues unless there is plentiful ram available.


varilci (approved):
lgtm

PR #551, opened 2021-03-17
sys-apps/portage: fork stable

xmhd:
Fork sys-apps/portage stable version to apply patch which silences the deprecated profile check.


varilci (approved):
Works as intended -both on a clean chroot environment and on an unimportant Hetzner machine-.

PR #556, opened 2021-03-29
dev-libs/openssl: import 1.0.2 with CVE patches from Ubuntu

xmhd:
1.0.2 has CVE's which require patching. Said patches are from Ubuntu 16.04 which is still supported by Canonical.

Tested on ~amd64 Gentoo Linux machine.


xmhd:
@varilci

xmhd:
> Lgtm. Tested on a ~amd64 machine with FEATURES=test without issues.
> 
> @xmhd, we should do the same for openssl-compat-1.0.2u.

shall I add the openssl-compact-1.0.2u fixes to this PR?

alip:
openssl-compat tested on a ~amd64 machine without issues.

PR #557, opened 2021-03-30
clean-up to fix QA

xmhd:
QA is failing due to old KEYWORDS, USE and DEPEND from some older imported packages. Clean those up with this PR.


varilci (approved):
looks good

PR #567, opened 2021-04-15
dev-perl/Ruby-VersionManager: inherit vcs-snapshot eclass.

monsieurp:
If you find yourself tinkering with the directory name in the `src_unpack` phase, simply inherit the `vcs-snapshot` eclass and enjoy life instead.

@xmhd @varilci


xmhd (approved):
rep++

PR #576, opened 2021-04-19
dev-python/* clean up

xmhd:
as per title


dev-python/EasyProcess/EasyProcess-0.2.5.ebuild
```diff
@@ -3,16 +3,19 @@
```
varilci, line 3: no 2021?

varilci (approved):
bombs away

dev-python/EasyProcess/EasyProcess-0.2.5.ebuild
xmhd: good catch, rep++

PR #583, opened 2021-04-28
net-analyzer/luigi: EAPI bump and PV bump

xmhd:
as per title.

ebuild could still do with some further improvement, e.g. conf file and such.


varilci (changes requested):
You left ebuild of `luigi-3.0.2` untouched but deleted it from Manifest, hence pkgcheck is failing.

xmhd:
> You left ebuild of `luigi-3.0.2` untouched but deleted it from Manifest, hence pkgcheck is failing.

fixed fixed fixed

also added missing RDEPEND

varilci (approved):
looks good now

Issue #596, opened 2021-06-03
Create Aerospike Prometheus Exporter ebuild


lemanchester:
Aerospike's Monitoring Stack extracts operational metrics from Aerospike database clusters for visualization and analysis in Prometheus and Grafana monitoring software.

source: https://github.com/aerospike/aerospike-prometheus-exporter 

Overview of the Monitoring stack: 

https://docs.aerospike.com/docs/tools/monitorstack/index.html

lemanchester:
some reference on how to handle go ebuild: 

https://github.com/adjust/gentoo-overlay/commit/1cf88f8592dbd0a1550648be1becaac34902334e#diff-f83de6767298b537397209ad18d608b208193ec6a8ed60eb87543a70ef1acaad

PR #597, opened 2021-06-03
Aerospike prometeus exporter init

lemanchester:
Solves #596

In progress, I will need some time to update to use `inherit go-module user` and set the `EGO_SUM`


net-analyzer/aerospike-prometheus-exporter/aerospike-prometheus-exporter-1.2.0.ebuild
```diff
@@ -0,0 +1,56 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
lemanchester, line 4: that should be 7

xmhd:
So like we discussed on slack, you'll want to switch eclass inherit from golang-base to golang-module, see the changes I made to statsd_exporter package in this commit https://github.com/adjust/gentoo-overlay/commit/1cf88f8592dbd0a1550648be1becaac34902334e

dev-go/get-ego-vendor package in gentoo will give you a tool(s) to parse go.mod and go.sum from the aerospike-prometheus-exporter repo, and give you EGO_SUM=(..) for you to add to the ebuild. You'll need to change src_compile to:

```
src_compile() {
        go build
}
```

go-module eclass is really nice and makes packaging software written in golang _way_ easier than it used to be with golang-base etc.

The other thing to be aware of is the enewuser / enewgroup stuff in pkg_setup. We are in the process of migrating packages to use the new setup of acct-user/foo + acct-group/foo packages, see the GLEP-81 migration PR. These ebuilds are pretty much all the same other than different user id + group id. If you need any help / advice on this, we are happy to help.

With regards to the init script, I would prefer multiple people review this as it is a critical piece of system software. These scripts should all be in posix shell and free of bashisms.
cc @alip @varilci @monsieurp ^^

The other thing on my mind right now is naming conventions and package categories - right now we have app-metrics/statsd_exporter and net-analyzer/prometheus-postgres_exporter, and in this PR there is net-analyzer/aerospike-prometheus_exporter. Does anyone have an opinion on naming / package category? ideally these should all match up with each other.

xmhd:
I've migrated the ebuild to use go-module eclass and added the relevant information. The ebuild will compile the software successfully now, but I haven't tried running src_install yet - it probably needs some work. The init script will also need work + review.

I would like to propose migrating all *-exporter packages to the app-metrics category to keep these in line with upstream Gentoo. Does anyone have any complaints about this?

xmhd:
I've moved this to the app-metrics category, and it is ready for testing.

@alip can you test/review please? for me it builds on Gentoo ~amd64.

xmhd (approved):

PR #598, opened 2021-06-03
dev-db/bagger-exports: tweak required schaufel ver

xmhd:
as per title.

@65278 is this ok? I see bagger-exports-0.0.7 requires schaufel-0.7, and so on for other vers. schaufel-0.9 doesn't exist now and is leaving a qa error.


xmhd:
closed, PR won't work. schaufel-0.9 incoming soon (tm)

PR #608, opened 2021-06-29
flux-bin: add pkg v0.15.3

mcharriere:


xmhd:
I fixed up flux-bin. My observations are:

- I tried switching this to a from source build with no luck. The source tarball distributed by flux does not contain some manifests which need to be fetched at build time - this doesn't work in packaging where sandboxing is applied.
- No man pages are distributed with the pre built binaries (at least from what I could see?)

xmhd (approved):

PR #609, opened 2021-06-30
virtual/adjust-infrastructure: small tweaks

xmhd:
Small tweaks for the ebuild as a whole.

e.g. only depend on dev-perl/Rex-foo-bar if app-admin/rex is in use.

Default to app-admin/trex.


varilci (approved):
lgtm

PR #610, opened 2021-07-02
sys-cluster/flux-bin: bump version

mcharriere:


xmhd (approved):
good catch, my mistake :D

PR #615, opened 2021-07-09
dev-lang/{elixir,erlang}: rescue from ::gentoo

xmhd:
as per title. required for AB1

tested && confirmed working on ~amd64


varilci (approved):
Looks good

Issue #559, opened 2021-03-31
GLEP 81 tracker


xmhd:
[https://www.gentoo.org/glep/glep-0081.html](url)

GLEP 81 in Gentoo Linux brought in management of users + groups by dedicated packages. Packages should be transitioned to this as opposed to continuing to use enewuser and enewgroup.

This issue shall serve as the tracker for discussing migration and tracking progress.

Some of these packages will already be assigned a user + group ID upstream at Gentoo. We should stay 1:1 with those where possible. For packages which are not, we should check what other Linux distributions / BSDs are assigning.

note: some of these packages can most likely be purged from the overlay which will make our lives easier.

Packages in ::adjust which are currently affected and require migration:

- [X] app-admin/consul-bin: GID 411 UID 411 (Gentoo)
- [X] app-admin/schaufel: GID 997 UID 997
- [X] app-metrics/prometheus: GID 430 UID 430 (Gentoo)
- [X] app-metrics/statsd-explorer
- [X] app-misc/elasticsearch: GID 183 UID 183 (Gentoo)
- [ ] app-misc/go-carbon
- [X] app-misc/karmabot-bin GID 114 UID 114 - most commonly recurring ID in our infrastructure
- [X] app-misc/logstash: GID 270 UID 270 (Gentoo)
- [X] dev-db/aerospike-server-enterprise GID 118 UID 118 - most commonly recurring ID in our infrastructure
- [X] dev-db/clickhouse: GID 998 UID 998 - most commonly recurring ID in our infrastructure
- [X] dev-db/metabase-bin GID 999 UID 999 - most commonly recurring ID in our infrastructure
- [X] net-analyzer/clickhouse_exporter GID 109 UID 109 - most commonly recurring ID in our infrastructure
- [X] net-analyzer/nmis GID 119 UID 119 - most commonly recurring ID in our infrastructure
- [X] net-analyzer/postgresql_exporter
- [X] net-analyzer/prometheus GID 430 UID 430 (Gentoo)
- [X] net-analyzer/prometheus-postgres_exporter
- [X] net-analyzer/zabbix GID 329 UID 329 (Gentoo)
- [X] sys-cluster/ceph: GID 267 UID 267 (Gentoo)
- [X] sys-cluster/chronos-bin GID 2000 UID 2000 - most commonly recurring ID in our infrastructure
- [X] sys-cluster/flink-bin GID 5013 UID 5013 - most commonly recurring ID in our infrastructure
[...]

PR #617, opened 2021-07-14
GLEP-81 migration

xmhd:
GLEP 81 migrations for various packages

See: #559

GID + UID are an initial stab, based off what ID's these groups/users have on our existing systems.

Don't merge this until it has been reviewed by multiple people - there may be GID/UID changes required, or even just set to -1.


xmhd:
Confirmed good as per @alip :+1: 
- aerospike
- zabbix
- logstash
- prometheus

xmhd:
I've looked over all of this, any I'm unsure about I have set to -1 for gid + uid (we can change these later - this just keeps current behaviour but migrates to new packages for handling gid + uid).

Please review when possible @alip @varilci @monsieurp

dev-db/metabase-bin/metabase-bin-0.34.0.ebuild
```diff
@@ -1,10 +1,8 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2019 Gentoo Authors
```
varilci, line 1: 2021?

xmhd: fixed

xmhd:
> It looks good to me, we may want one more review from Patrice before merging this. -that's why I am not giving a direct approval-
> 
> I can provide an approval if desired.

there is no rush, we can wait on patrice and alip.

PR #621, opened 2021-07-26
eclass/python-utils-r1: fork eclass for PYTHON_IMPL tweaks

xmhd:
tldr; upstream is aggressive in removing compatible PYTHON_IMPL which leaves many of our machines without a working Python environment..... so now we fork python-utils-r1 and tweak PYTHON_IMPL to fix that.


varilci (approved):
lgtm

xmhd:
deployed && tested && confirmed working on unbotify-rts hosts

PR #628, opened 2021-08-10
dev-db/postgresql: add hugelwlock useflag

65278:
Work in progress: Finding the right tuning to avoid buffer mapping locks on seriously huge postgres instances.


65278:
We've decided that 512 (1 << 9)is a good starting value and we'll iterate on it later.

xmhd (approved):

PR #629, opened 2021-08-11
Update apache-arrow + thrift to latest stable releases

xmhd:
- apache-arrow bumped to v4.0.0
- thrift bumped to v0.14.1

apache-arrow ebuild rev bumped for slight refactoring.

I have tested on ~amd64 and it seems to work. I have retained the old version of arrow so that we can downgrade it on the AB testing servers if it is borked.


monsieurp (changes requested):
* https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-7-variables make sure to add BDEPEND as it looks like the variable is now required with EAPI 7
* the thrift ebuild leverages slot operators whereas the apache arrow ebuild doesn't

xmhd:
>     * https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-7-variables make sure to add BDEPEND as it looks like the variable is now required with EAPI 7
> 
>     * the thrift ebuild leverages slot operators whereas the apache arrow ebuild doesn't

the latter should be looking ok now (disregard the 0.x ebuild, that is legacy).

monsieurp (approved):

PR #631, opened 2021-08-20
calicoctl

mcharriere:
Add calicoctl taken from https://packages.gentoo.org/packages/net-misc/calicoctl and update it to 3.19.2

- net-misc/calicoctl: initial import
- net-misc/calicoctl: bump version


mcharriere:
hashes in EGO_VERSION need to be updated.

xmhd:
migrated existing ebuild to use go-module eclass.

tested on ~amd64

PR #637, opened 2021-08-27
sys-cluster/mesos: create eclass and inherit it in the ebuild.

monsieurp:
create mesos.eclass and inherit it in the ebuild.


monsieurp:
Note: tested on esh-ab-worker-7.adjust.com.

varilci (approved):
lgtm

PR #644, opened 2021-09-06
[WIP]: */*: QA fixes.

monsieurp:
WIP: more commits to come.


app-metrics/postgresql_exporter/postgresql_exporter-0.1.1.ebuild
```diff
@@ -21,8 +21,6 @@ IUSE=""
 RDEPEND="
 	acct-group/postgresql_exporter
 	acct-user/postgresql_exporter
-
-	!net-analyzer/prometheus-postgres_exporter
```
xmhd, line 25: this needs to remain in the postgresql_exporter ebuild, tldr; prometheus-postgres_exporter is the upstream software whereas postgresql_exporter is adjust's own software, and they cannot co-exist.

I did mean to leave a comment in the ebuild for this, but evidently I forgot.

PR #648, opened 2021-09-08
dev-db/postgresql: delete pg-legacytimestamp in use_enable.

monsieurp:
Closes: https://github.com/adjust/gentoo-overlay/issues/647
Package-Manager: Portage-3.0.20, Repoman-3.0.3


65278 (changes requested):
Now we have the same problem with integer-datetimes.
This configure flag is about disabling 64 bit ints for timestamp (and use double floats instead).

I've considered this, nobody in their right mind would enable this anymore, and I don't think it's worth supporting. Just delete the offending line altogether.

65278 (commented):
LGTM.
```2021-09-08T08:35:53 >>> dev-db/postgresql-9.6.22: 2′27″```

PR #649, opened 2021-09-13
sys-kernel/genkernel: implement logic for fetching squashfs image via network

xmhd:
As per title.

Genkernel supports squashfs images embedded in the initramfs (as this was designed for liveCD's), however there is currently no logic for fetching an image via network.

This PR adds a new parameter 'fetch_sqfs=foo" to be passed which will provide a uri to fetch the squashfs image from. startup_network is handled if this parameter is used, and there is a conditional added to the startup_network code for USE_SSHD, so that we don't call start_network twice.

### DIFF
--- a/sys-kernel/genkernel/files/genkernel-4.2.3-fetch_sqfs.patch
+++ b/sys-kernel/genkernel/files/genkernel-4.2.3-fetch_sqfs.patch
@@ -1,5 +1,5 @@
 diff --git a/defaults/linuxrc b/defaults/linuxrc
-index 6ede740..a485358 100644
+index 6ede740..1e36719 100644
 --- a/defaults/linuxrc
 +++ b/defaults/linuxrc
 @@ -179,6 +179,10 @@ do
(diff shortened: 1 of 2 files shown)


xmhd:
@monsieurp

sys-kernel/genkernel/files/genkernel-4.2.3-fetch_sqfs.patch
```diff
@@ -0,0 +1,63 @@
...
++		;;
+ 		# iSCSI
+ 		iscsi_initiatorname=*)
+ 			ISCSI_INITIATORNAME=${x#*=}
+@@ -615,6 +619,19 @@ then
+ 	start_iscsi
+ fi
+ 
++# optionally fetch a squashfs image from the network
++if [ "${USE_SQFS_FETCH}" = '1' ]
```
monsieurp, line 21:
```suggestion
+if [ "${USE_SQFS_FETCH}" eq 1 ]
```

monsieurp (changes requested):

sys-kernel/genkernel/files/genkernel-4.2.3-fetch_sqfs.patch
xmhd: I prefer the latter, but the former keeps to the convention used throughout genkernel and stands us the best chance of having this accepted upstream.

monsieurp:
As per our Slack chat, we should break the fetching and the writting out to ${LOOP} into two different steps.

sys-kernel/genkernel/files/genkernel-4.2.3-fetch_sqfs.patch
```diff
@@ -0,0 +1,63 @@
...
++then
++	start_network
++
++	good_msg "Fetching $SQFS_URI"
++
++	# CWD is /newroot
++	wget -O - $SQFS_URI > ${LOOP} && good_msg "Written to ${LOOP}"
++
++	unset SQFS_URI
```
monsieurp, line 30:
```suggestion
+	# CWD is /newroot
+     sqfs_img=$(basename $SQFS_URI)
+	good_msg "Fetching $SQFS_URI ..."
+     wget -O - $SQFS_URI > "/newroot/$sqfs_img" && good_msg "Fetched $SQFS_URI successfully!"
+
+     good_msg "Writting $sqfs_uri to ${LOOP} ..."
+     cat "/newroot/$sqfs_img" > ${LOOP} && good_msg "$sqfs_img written to ${LOOP} successfully!"
+     unset SQFS_URI sqfs_img
```

monsieurp (changes requested):

monsieurp (approved):

PR #650, opened 2021-09-17
net-proxy/oauth2-proxy: initial configuration

artemyevav:
There's no official gentoo package for `oauth2-proxy`. This PR wraps official github binaries so that `oauth2-proxy` can be installed on gentoo.
Oauth2 Proxy repo: https://github.com/oauth2-proxy/oauth2-proxy


xmhd:
lgtm , if you want to do from source builds for software written in go, utilise the go-module eclass (see the app-metrics/statds_exporter package for example). EGO_SUMS is output from running get-ego-vendor in the cloned repository.

artemyevav:
@xmhd go class was my first attempt but I was stuck fixing dependencies in 3rd party packages. It just doesn't work.

PR #651, opened 2021-09-20
basic nodejs eclass, + dev-nodejs package category

xmhd:
We need to have a NodeJS package available for configurable-http-proxy, so we can either install manually using npm, or use some publicly existing work for packaging NodeJS stuff in Gentoo.

This PR merges an eclass from https://github.com/Jannis234/jm-overlay with some tweaks, and adds a dev-nodejs package category to profiles/categories.

This PR also adds a dev-nodejs/configurable-http-proxy package which appears to work at first glance (on my ~amd64 box).

There is another approach for packaging NodeJS/npm stuff being done at this overlay:
https://github.com/orsonteodoro/oiledmachine-overlay/blob/master/eclass/npm-utils.eclass
https://github.com/orsonteodoro/oiledmachine-overlay/blob/master/eclass/npm-secaudit.eclass

It seems like a more in-depth approach but also seems to be a bit more heavyweight (I don't think we'll have much packaged using this).

There is also this approach here https://github.com/Alessandro-Barbieri/node-overlay/tree/main/eclass

wdyt? @monsieurp @varilci


rashamalek:
@monsieurp @alip is this PR still relevant? can we merge it?

PR #653, opened 2021-09-27
net-proxy/oauth2-proxy: migrate to go-module eclass

xmhd:
as per title.


xmhd:
tested on ~amd64

varilci (approved):
lgtm, tested with bare amd64 image

PR #654, opened 2021-09-27
dev-go/centrifugo: convert to using go-module eclass

xmhd:
as per title.

tested on ~amd64


varilci (approved):
lgtm, tested with bare amd64 image. Minor point:

> MissingLicense: version 2.7.0-r1: no license defined

xmhd:
> lgtm, tested with bare amd64 image. Minor point:
> 
> > MissingLicense: version 2.7.0-r1: no license defined

fix'd

PR #675, opened 2021-12-03
acct-user/github_runner: Add home to user

licebmi:
Some applications (like vagrant) require a home directory to be
available.
@artemyevav @xmhd

### DIFF
--- a/acct-user/github_runner/github_runner-0.ebuild
+++ b/acct-user/github_runner/github_runner-0.ebuild
@@ -7,5 +7,6 @@ inherit acct-user
 
 ACCT_USER_ID=-1
 ACCT_USER_GROUPS=( ${PN} )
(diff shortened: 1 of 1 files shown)


artemyevav:
actions-runner is inttalled to `/opt` https://github.com/adjust/gentoo-overlay/blob/ef3b48bde37d8daca006d52b6e3c8aa4a2383eee/dev-util/actions-runner/actions-runner-2.284.0.ebuild#L18
Makes it sense to set that as a homedir?

licebmi:
@artemyevav I was planning to use that first, but I decided against since the runner will populate it's files, and I didn't want to risk file name collisions.

artemyevav:
OK then, LGTM. Awaiting for @xmhd tho.

xmhd (approved):

PR #677, opened 2021-12-06
sys-kernel/debian-sources: initial check-in

xmhd:
new kernel ebuild which will handle:

- fetching of kernel sources
- fetching of debian patches
- generation of debian kconfig
- tweaking of kconfig based on USE flags
- kernel compilation
- initramfs creation
- external kernel modules rebuild via subslot(s) (same beahviour as sys-kernel/gentoo-kernel in ::gentoo)
- initramfs rebuild in pkg_postinst (i.e. if binpkg is emerged from binhost, new initramfs will be created with host specific info e.g. udev rules, mdadm.conf)

virtual/linux-sources changes are to allow sys-kernel/debian-sources to satisfy this virtual.

eclass/linux-mod changes are to allow USE=build-kernel in place of USE=dist-kernel.

Has been tested for months on various hosts and engineers local machines.


xmhd:
TODO: 
- mount-boot_pkg_pretend doesn't seem to work properly?

then this can be merged.

xmhd:
mount-boot and check-reqs eclass are fubar, we can perform these checks in Rex so I've removed for now

PR #680, opened 2021-12-11
runners-supervisor: initial config

artemyevav:
Initial config of `dev-util/runners-supervisor` package.
This package implements autoscaling solution on the gentoo server without using k8s or aws (see https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#recommended-autoscaling-solutions). It's a bash script that creates, configures, updates and monitors self-hosted ephemeral github runners.


artemyevav:
@xmhd @lemanchester any feedback would be appreciated

PR #682, opened 2021-12-21
dev-db/aerospike-server-enterprise: Fix 4.4.0.9

lemanchester:
related to https://github.com/adjust/gentoo-overlay/pull/681


lemanchester:
I'm not sure if works

PR #689, opened 2022-01-24
Aerospike split

roa:
Split the aerospike ebuild in 2 parts:

**aerospike-server-enterprise**

The server binary and the files needed to start it. For this I created the `-r1` ebuild for version `4.5.1.8`. All `-r1` would then have the split.

**aerospike-server-tools**

The aerospike tools:

```
asbackup
asbenchmark
aql
asvalidation
asadm
asloglatency
asinfo
asrestore
remove_python_path
asloader
add_python_path
```

These depend on python3, so we don't need to depend on python2 anymore.


xmhd:
I would like to take this opportunity to bump the EAPI to 7 on these ebuilds.

cc @monsieurp

roa:
I changed the ebuilds to EAPI7

PR #696, opened 2022-02-02
aerospike-zabbix-plugin: Additional wrapping scripts

zegeek:
Adding these additional wrapping scripts will allow us to generate different items in zabbix with the same externalscript :smile:


xmhd:
change:

```
#!/bin/bash
```

to

```
#!/usr/bin/env bash
```

cheers

zegeek:
Done :D

PR #699, opened 2022-02-07
change ownership of log directory + edit path in default openrc script

zegeek:
After emerge : the log files directory is not owned by pgbouncer:postgres and the path pointed at by the openrc init script is dynamically changing according to the openrc unit name (the directory should be static, but the log file itself should be dynamic)


xmhd:
add:

```
acct-group/postgres
```

to RDEPEND

and remove:

```
diropts -o pgbouncer -g postgres -m 0750
```

xmhd (approved):

PR #700, opened 2022-02-08
add victoria-metrics support for cluster build

zegeek:
adding the necessary openrc services to support multiple vminsert, vmselect and vmstorage running on the same node.
Please do not merge, it's a work in progress


mcharriere (commented):
could you drop the `-r1` from the ebuild name?

PR #705, opened 2022-02-22
www-servers/uwsgi: remove patchlevel

alexie-smirnoff:
Remove patch level from www-servers/uwsgi to match the upstream.
fixes https://github.com/adjust/infrastructure/issues/15486


varilci (approved):
lgtm

PR #710, opened 2022-03-01
Add support for nocairo carbonapi build

zegeek:
Carbonapi offers a graphical support for generating graphs based on cairo lib. We do not use this as we interface carbon api all the time with grafana instead. 
This use flag will allow us to build carbonapi without cairo support from portage layer.
ref: https://github.com/go-graphite/carbonapi/blob/main/BUILD.md#build-instructions


xmhd:
is nocairo an additional target to the default? or does running 'make nocairo' build the entire carbonapi as expected?

zegeek:
@xmhd make nocairo will build the entire carbonapi as expected, just without cairo support ( so it won't render any graphic, which is fine by me)

xmhd:
@zegeek can you rebase this? ping me on slack after and I will merge.

xmhd (approved):

Issue #635, opened 2021-08-24
Create jmx_exporter and dependencies ebuild


oziey88:
JMX exporter extracts operational metrics from kafka clusters for visualization and analysis in Prometheus and Grafana monitoring software.

source: jmx_exporter https://github.com/prometheus/jmx_exporter

ref: https://github.com/adjust/infrastructure/issues/13469

oziey88:
https://github.com/adjust/infrastructure/issues/13469

oziey88:
we will use jmx_exporter as java agent on kafka, no need to create a saperate package for this.  
closing

PR #714, opened 2022-03-11
app-metrics/jmx_exporter: first stab at ebuild

xmhd:
As per title.

#635 

@oziey88 @lemanchester the home page for jmx_exporter tells me that it comes in two parts:

- java agent
- http server

This package currently only contains the java agent. Do we require anything further?

I have included sample configuration files in files/example-configs. A configuration file for use at Adjust will need to be created.

I'll create a shell script which will server as a launcher for the jar file which will also be installed to /opt/jmx_epxorter and symlinked to /usr/bin/jmx_exporter.


xmhd:
my google-fu tells me that the http server portion is indeed required. I'll add that to this package soon (tm).

PR #720, opened 2022-03-22
[NOMERGE] app-shells/bash: ebuild for static binary packages

alexie-smirnoff:


alexie-smirnoff:
Switch to this branch for building with an old `glibc` when needed.

PR #723, opened 2022-03-30
Schaufel libconfig parser bug

65278:
Libconfig 1.5 has a bug concerning its grammar (no comma before empty element in a list). 
To force all servers to bump libconfig on schaufel build, I've updated the runtime dep.


zegeek (approved):
We tested it and everything works fine that way.

PR #727, opened 2022-04-28
aerospike: add missing mountpoint detection in pre checks

zegeek:
To prevent starting aerospike without it's index-on-disk mountpoints, we add the following checks in the init script.
related to adjust/infrastructure#16215

### DIFF
--- a/dev-db/aerospike-server-enterprise/files/aerospike.init
+++ b/dev-db/aerospike-server-enterprise/files/aerospike.init
@@ -46,6 +46,13 @@ start_pre() {
 	set_shmmax
 	ulimit -n 100000
 	if [ -n $LD_PRELOAD ]; then export LD_PRELOAD; fi
+	#checking index-on-disk mount points
+	for m in $(awk '/mount / { print $2}; ' /etc/aerospike/aerospike.conf); do
+            if ! grep -qs $m /proc/mounts; then
+                eerror "mountpoint $m is missing"
+                return 1
+            fi
+        done
 }
 
 start_post() {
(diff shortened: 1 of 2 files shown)


monsieurp (changes requested):
Could you pls split the code across several lines instead of using a one-liner? Thanks!

dev-db/aerospike-server-enterprise/files/aerospike.init
```diff
@@ -46,6 +46,11 @@ start_pre() {
 	set_shmmax
 	ulimit -n 100000
 	if [ -n $LD_PRELOAD ]; then export LD_PRELOAD; fi
+	#checking index-on-disk mount points
+	for mount in `pcregrep -M "index-type flash \{[\sa-z\-/0-9]+" ${config_file} | grep -E "mount\s+" | awk '{print $2}'`
+	do 
+		df $mount >/dev/null 2>&1 || { eerror "mountpoint $mount is missing"; return 1;}
+	done
```
monsieurp, line 53:
```suggestion
        for m in $(awk '/mount / { print $2}; ' /etc/aerospike/aerospike.conf); do
            if ! grep -qs $m /proc/mounts; then
                eerror "mountpoint $m is missing"
                return 1
            fi
        done
```

dev-db/aerospike-server-enterprise/files/aerospike.init2
```diff
@@ -42,6 +42,11 @@ start_pre() {
 	set_shmmax
 	ulimit -n 100000
 	if [ -n $LD_PRELOAD ]; then export LD_PRELOAD; fi
+	#checking index-on-disk mount points
+	for mount in `pcregrep -M "index-type flash \{[\sa-z\-/0-9]+" ${config_file} | grep -E "mount\s+" | awk '{print $2}'`
+	do 
+		df $mount >/dev/null 2>&1 || { eerror "mountpoint $mount is missing"; return 1;}
+	done
```
monsieurp, line 49: Same change.

monsieurp (changes requested):

monsieurp (approved):

PR #730, opened 2022-05-09
dev-util/sonarqube-developer

licebmi:
Sonarqube developer server package.


LehningerAdjust:
@monsieurp Please review.

monsieurp:
CI is broken unfortunately so please ignore for now.

dev-util/sonarqube-developer/files/exit-codes.patch
```diff
@@ -0,0 +1,36 @@
+--- a/bin/linux-x86-64/sonar.sh
++++ b/bin/linux-x86-64/sonar.sh
+@@ -439,6 +439,7 @@ start() {
+         echo "Started $APP_LONG_NAME."
+     else
+         echo "Failed to start $APP_LONG_NAME."
++	exit 1
```
monsieurp, line 7: Missing space here:
```suggestion
         exit 1
```

dev-util/sonarqube-developer/sonarqube-developer-8.9.8.54436.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	"${FILESDIR}/pidfile.patch"
+	"${FILESDIR}/exit-codes.patch"
+	"${FILESDIR}/configuration.patch"
+)
+
+inherit epatch
+
+src_unpack() {
+	unpack ${P}.zip
+	mv sonarqube-${PV} ${P}
```
monsieurp, line 35:
```suggestion
        mv sonarqube-${PV} ${P} || die
```

monsieurp (changes requested):

dev-util/sonarqube-developer/sonarqube-developer-8.9.8.54436.ebuild
licebmi: Is this required?
I mean the zipfile is verified with checksum by portage, and mv is an atomic operation which just migrates an inode. Plus the next step will fail if ${P} doesn't exists.

monsieurp (approved):

dev-util/sonarqube-developer/sonarqube-developer-8.9.8.54436.ebuild
monsieurp: Hey! Yes it's better to add `|| die` statements to make sure the file is indeed renamed.

PR #733, opened 2022-05-16
Pgbackrest TLS server

65278:
Pgbackrest added a TLS server for remote backups in 2.37.

I found this feature useful, so I've written some glue for it.

### DIFF
--- a/dev-db/pgbackrest/files/pgbackrest.initd
+++ b/dev-db/pgbackrest/files/pgbackrest.initd
@@ -47,11 +47,8 @@ start()
 {
 	ebegin "Starting ${SVCNAME}"
 	checkconfig || return 1
-	local clients=""
-	for client in ${SERVER_AUTH/;/ }
-	do
-		clients="$clients --tls-server-auth=$client"
-	done
+	local clients="$(echo --tls-server-auth=$SERVER_AUTH |\
+		sed 's/;/ --tls-server-auth=/g')"
 
 	start-stop-daemon --start --exec /usr/bin/pgbackrest \
 		--user postgres --group postgres \
--- a/dev-db/pgbackrest/files/pgbackrest.initd
+++ b/dev-db/pgbackrest/files/pgbackrest.initd
@@ -45,7 +45,8 @@ checkconfig() {
 
 start()
 {
-	checkconfig || eend 1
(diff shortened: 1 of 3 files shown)


dev-db/pgbackrest/files/pgbackrest.initd
```diff
@@ -0,0 +1,74 @@
...
+	then
+		eerror "${SVCNAME} has detected a syntax error in your configuration file: "
+		pgbackrest info
+	fi
+	return $ret
+}
+
+start()
+{
+	checkconfig || eend 1
```
alip, line 47: Note, `eend` is an informational output command and as such does not cause early return here which seems to be what's intended. Also you usually want to pair `eend` with `ebegin`.

65278: You're totally right. `eend` could only end the program if it execs and exits. My brain took a leap there.

Also, `ebegin` is nice to have.

dev-db/pgbackrest/files/pgbackrest.initd
```diff
@@ -0,0 +1,78 @@
...
+	fi
+	return $ret
+}
+
+start()
+{
+	ebegin "Starting ${SVCNAME}"
+	checkconfig || return 1
+	local clients=""
+	for client in ${SERVER_AUTH/;/ }
```
alip, line 51:
```
 * QA Notice: shell script appears to use non-POSIX feature(s):
 *    possible bashism in /etc/init.d/pgbackrest line 51 (${parm/?/pat[/str]}):
 *      for client in ${SERVER_AUTH/;/ }
```

65278: Alright. I've addressed this. The code is now objectively worse. 
And now an out of place rant: Yes, you can replace /bin/sh with another shell, and openrc-run will invoke that. Since gentoos ebuilds depend on bash, I see no point in openrc-run not depending on bash.

alip: This discussion is at least 15 years old and has seen many flamewars in -dev. I agree with you but since it's a QA issue it's a good idea to fix it imho. Thank you.

alip (approved):

PR #749, opened 2022-07-12
sys-kernel/debian-sources: version bump.

monsieurp:
cc @65278


65278 (approved):
LGTM. For everybody else:
It's important we don't regress below 5.17.6 because of a cephfs kernel bug.

monsieurp:
Thanks for the heads up @65278 👍 We'll keep it in mind for the future.

PR #760, opened 2022-08-22
ebuild file for grafana 9.1.0

ibra-adjust:
Description
we need to implement ebuild for Grafana 9.1.0, to setup Grafana HA with the last Grafana version 9.1.0 because there is a security issue [CVE-2022-31097](https://github.com/grafana/grafana/security/advisories/GHSA-vw7q-p2qg-4m5f) in the previous versions, but the last version on gentoo packages is 9.0.2.


ibra-adjust:
The package is added to the gentoo packages

PR #761, opened 2022-08-24
bagger: bump to 0.7

65278:
Not much to say.
Very many important bugfixes to bagger itself, bagger-data and bagger-master are bumped in name only.


zegeek (approved):
looks good ;)

PR #770, opened 2022-09-19
files/vmagent.{confd,initd}: create generic init files for vmagent

zegeek:
To expand the victoriametrics cluster and in order to allow vm to scrape metrics exports à la Prometheus, we need to deploy vmagents that will scrape the /metrics endpoints and write to victoriametrics. 
This PR adds the init/conf file of a generic vmagent, that we will provision using TRex/cmdb.
related to adjust/infrastructure#17556


dev-db/victoria-metrics/files/vmagent.initd
```diff
@@ -0,0 +1,46 @@
...
+
+httpListenAddr="${HTTP_LISTEN_ADDR:-:8429}"
+
+
+name="VictoriaMetrics ${RC_SVCNAME}"
+
+pidfile="/run/victoria-metrics/${RC_SVCNAME}"
+
+command="/usr/bin/vmagent"
+command_args="-httpListenAddr=${httpListenAddr} -envflag.enable -envflag.prefix=${RC_SVCNAME}_"
```
monsieurp, line 29: `-envflag.prefix=${RC_SVCNAME}_"`
there's an extra `_` here is it normal?

dev-db/victoria-metrics/files/vmagent.initd
```diff
@@ -0,0 +1,46 @@
...
+	need local
+	use net
+}
+
+start_pre() {
+	remote_write_tmpdir=${RC_SVCNAME}_remoteWrite_tmpDataPath
+	config_file=${RC_SVCNAME}_promscrape_config
+	checkpath -d -m 0755 -o "${command_user}" $(dirname ${output_log})
+    checkpath -d -m 0660 -o "${command_user}" ${!remote_write_tmpdir}
+    checkpath -f -m 0440 -o "${command_user}" ${!config_file}
```
monsieurp, line 45: indentation

dev-db/victoria-metrics/files/vmagent.initd
```diff
@@ -0,0 +1,46 @@
...
+export ${RC_SVCNAME}_memory_allowedBytes="${MEMORY_ALLOWEDBYTES:-0}"
+export ${RC_SVCNAME}_memory_allowedPercent="${MEMORY_ALLOWEDPERCENT:-60}"
+export ${RC_SVCNAME}_promscrape_cluster_memberNum="${MEMBER_NUM:-0}"
+export ${RC_SVCNAME}_promscrape_cluster_membersCount="${MEMBER_COUNT:-0}"
+export ${RC_SVCNAME}_promscrape_cluster_name="${CLUSTER_NAME:-}"
+export ${RC_SVCNAME}_promscrape_config="${PROMSCRAPE_CONFIG:-/etc/vmagent/vmagent.yml}"
+export ${RC_SVCNAME}_promscrape_config_strictParse="${STRICT_PARSE:-true}"
+export ${RC_SVCNAME}_remoteWrite_url="${REMOTE_WRITE_URL:-http://localhost/insert/prometheus/api/v1/write}"
+export ${RC_SVCNAME}_remoteWrite_tmpDataPath="${TMP_DATA_PATH:-/var/tmp/vmagent}"
+export ${RC_SVCNAME}_remoteWrite_maxDiskUsagePerURL="${MAX_DISK_USAGE:-200GB}"
```
monsieurp, line 19: a couple of questions here:
* why do you export those variables?
* you should set a default value in the confd file instead of using the `${FOOBAR:-value}` pattern, it makes more sense IMHO

monsieurp (changes requested):

dev-db/victoria-metrics/files/vmagent.initd
zegeek: yes, because the environment variables will end up having the name of the service as a prefix, like  `vmagent_1_<something_here>`

dev-db/victoria-metrics/files/vmagent.initd
zegeek: - exposing the variables is how we configure victoriametrics components, either by cmdline flags, or by having these env variables and passing the `-envflag.enable` option to the daemon. Taken from vmagent documentation:
```
 -envflag.enable
     Whether to enable reading flags from environment variables additionally to command line. Command line flag values have priority over values from environment vars. Flags are read only from command line if this flag isn't set
```
- based on other init scripts, usually the defaults are defined in the init script for example: 
https://github.com/adjust/gentoo-overlay/blob/bb63964cc1fbfd3f8c280403ac7e5e864f7f7daf/app-admin/schaufel/files/schaufel.initd#L3-L8
or 
https://github.com/adjust/gentoo-overlay/blob/bb63964cc1fbfd3f8c280403ac7e5e864f7f7daf/app-admin/filebeat/files/filebeat.initd#L5-L10

monsieurp: Upon starting a service, OpenRC will source its conf.d file first and pick up values from there. That's why I think setting defaults values in the init script is wrong and actually redundant.

I understand but it's not the right way to do it and we should definitely fix the schaufel ebuilds as well. See the mesos conf.d/init.d files here: https://github.com/adjust/gentoo-overlay/tree/master/sys-cluster/mesos/files

monsieurp (changes requested):

monsieurp (approved):

PR #771, opened 2022-09-19
aerospike-prometheus-exporter: fix ebuild and init script

zegeek:
Ebuild had some issues that didn't allow proper installation + missing ownership of certain directories change. 
Init script had a minor issues that prevented from starting the daemon.
related to adjust/infrastructure#17556

### DIFF
--- a/app-metrics/aerospike-prometheus-exporter/files/aerospike-prometheus-exporter.initd
+++ b/app-metrics/aerospike-prometheus-exporter/files/aerospike-prometheus-exporter.initd
@@ -8,13 +8,12 @@ pidfile="/run/${SVCNAME}/${SVCNAME}.pid"
 user="prometheus-exporter"
 
 command="/usr/bin/${SVCNAME}"
-command_args="${command_args}"
+command_args="${command_args} --config /etc/aerospike-prometheus-exporter/ape.toml"
 command_background="true"
 command_user="${user}"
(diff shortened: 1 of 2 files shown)


app-metrics/aerospike-prometheus-exporter/files/aerospike-prometheus-exporter.initd
```diff
@@ -8,23 +8,22 @@ pidfile="/run/${SVCNAME}/${SVCNAME}.pid"
...
 
 start_pre() {
-	if [[ -z ${DATA_SOURCE_NAME} ]]; then
-		ewarn "DATA_SOURCE_NAME not set"
-		exit 1
-	fi
+	#if [[ -z ${DATA_SOURCE_NAME} ]]; then
+	#	ewarn "DATA_SOURCE_NAME not set"
+	#	exit 1
+	#fi
```
monsieurp, line 27: pls get rid of those lines, we can still look into the repo's history if we need this code block again

monsieurp (changes requested):

app-metrics/aerospike-prometheus-exporter/files/aerospike-prometheus-exporter.initd
zegeek: yeah that makes sense

monsieurp (approved):

PR #778, opened 2022-11-01
sys-apps/portage: version bump to v3.0.38

alip:
Package-Manager: Portage-3.0.38, Repoman-3.0.3


alip:
- Bump portage to latest stable
- Drop the unneeded soname patch (already in upstream)
- pkgcheck failures are not related.

PR #782, opened 2022-11-17
net-misc/quagga: new package

monsieurp:


net-misc/quagga/quagga-1.2.4.ebuild
```diff
@@ -0,0 +1,137 @@
...
+You have to create config files in /etc/quagga before
+starting one of the daemons.
+
+You can pass additional options to the daemon by setting the EXTRA_OPTS
+variable in their respective file in /etc/conf.d"
+
+src_prepare() {
+	# Classless prefixes for BGP
+	# http://hasso.linux.ee/doku.php/english:network:quagga
+	use bgpclassless && eapply -p0 "${DISTDIR}/${CLASSLESS_BGP_PATCH}"
```
alip, line 59: Since this patch is no longer available, I think it's a good idea to drop the bgpclassless USE flag altogether.

net-misc/quagga/quagga-1.2.4.ebuild
```diff
@@ -0,0 +1,137 @@
...
+
+	eapply "${PATCHES[@]}"
+	eapply_user
+	eautoreconf
+}
```
alip, line 64: which in turn means we can drop `src_prepare` and let default_src_prepare handle the PATCHES. The single patch we have there does not touch autotools so we don't need autoreconf here.

alip (approved):
lgtm, i have a small comment.

PR #794, opened 2022-12-09
cloud-utils: Initial version

licebmi:
Added support to cloud utils, required for packing custom images for MAAS.


dev-util/cloud-utils/cloud-utils-0.33.ebuild
```diff
@@ -0,0 +1,43 @@
...
+RDEPEND="
+	${PYTHON_DEPS}
+	sys-fs/dosfstools
+	dev-libs/libisoburn
+	sys-fs/lvm2
+	app-emulation/qemu
+	app-arch/gzip
+	net-misc/wget
+	sys-apps/util-linux
+	sys-fs/udev
```
alip, line 28: It's a better idea to depend on `virtual/udev` rather than `sys-fs/udev` as there are different alternatives of udev implementations.

alip (commented):
lgtm overall, see my small comment below.
In addition, please add a `metadata.xml` file to the ebuild directory.

alip (approved):

PR #805, opened 2023-01-12
dev-db/timescaledb: use cmake eclass (rather than cmake-utils)

alip:
Package-Manager: Portage-3.0.41, Repoman-3.0.3


rashamalek:
@alip can you recheck this PR and merge it?

PR #812, opened 2023-01-20
app-admin/sos/ create ebuild

a-cato:
This is my attempt to create an ebuild for sosreport following the gentoo docs.
Feel free to make changes or to discard the whole thing and make it properly if it's too wrong.
Thanks!


app-admin/sos/sos-4.4.ebuild
```diff
@@ -0,0 +1,24 @@
...
+inherit distutils-r1
+
+DESCRIPTION="A tool that collects configuration details from a Linux system"
+HOMEPAGE="https://github.com/sosreport/sos"
+SRC_URI="https://github.com/sosreport/sos/archive/refs/tags/4.4.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
```
zegeek, line 17: There should be here a 
```
BDEPEND="
     dev-python/setuptools
"

File: app-admin/sos/sos-4.4.ebuild [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+RDEPEND="
+        dev-python/pycodestyle
+        dev-python/coverage
+        dev-python/sphinx
+        dev-python/pyyaml
+        dev-python/setuptools
```
zegeek, line 23: Here there is one more runtime dependency it is called dev-python/python-magic, it is required [here](https://github.com/sosreport/sos/blob/76c27a82b2688d0fb9a8b6bd5ea9ffd1c896fde9/sos/utilities.py#L25-L39) other wise it will dump a warning each time.

zegeek (changes requested):
@a-cato

app-admin/sos/files/sos-4.4-setup-py.patch
```diff
@@ -0,0 +1,11 @@
+--- a/setup.py	2023-01-23 13:41:44.731270118 -0000
++++ b/setup.py	2023-01-23 13:41:49.351270033 -0000
+@@ -93,8 +93,6 @@
+                             'man/en/sos-mask.1', 'man/en/sos-help.1']),
+         ('share/man/man5', ['man/en/sos.conf.5']),
+         ('share/licenses/sos', ['LICENSE']),
```
zegeek, line 6: I think you have to delete the last comma of this line @monsieurp

monsieurp (approved):

zegeek (approved):

PR #821, opened 2023-04-05
ebuild for postgresql 15.2

ibra-adjust:
Description
we need to implement ebuild for postgreSQL 15.2 to be install in the shard-buffer machines.


alip (commented):
PostgreSQL-15.2 is already packaged on Gentoo, see here:
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-db/postgresql/postgresql-15.2.ebuild
Is there any reason why we're not using that ebuild? Any differences here that we can push to upstream?

PR #827, opened 2023-06-13
dev-db/victoria-metrics: version bump

monsieurp:


dev-db/victoria-metrics/victoria-metrics-1.91.2.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	newinitd "${FILESDIR}/${PN}.initd" "${PN}"
+	newconfd "${FILESDIR}/${PN}.confd" "${PN}"
+
+	if use cluster; then
+		#cluster components
+		local cluster_comps=(
+			vminsert
+			vmselect
+			vmstorage
+		)
```
zegeek, line 49: @monsieurp can we please include vmagent also here

PR #860, opened 2024-01-25
app-admin/schaufel: add missing replication option in init script

zegeek:
REPLICATION option was missing from being handled in the init script, it drove me crazy because i didn't assume at all that the issue was there.


ibra-adjust (approved):
LGTM

PR #862, opened 2024-02-09
schaufel_exporter: make hostname dynamic through options

zegeek:
Make presented hostname to prometheus passed through option


roa (approved):
lgtm

PR #865, opened 2024-02-28
app-portage/gentoolkit: Add ebuild

licebmi:
Shamelessly copy-pasted from gentoo portage. This version was removed
and new version requires a more recent version of sys-apps/portage


LehningerAdjust (approved):
LGTM

PR #868, opened 2024-03-19
net-misc/loki-promtail-bin add ebuild

ibra-adjust:
Add an ebuild for net-misc/loki-promtail-bin


net-misc/loki-promtail-bin/files/promtail.yml
sibuthomasmathew, line 1: The configuration file must be kept independent of the build

ibra-adjust: Could you please explain why it must be kept independent of the build ?

sibuthomasmathew: A lot of the configuration like file `positions` or `client.url` can change. So, we need to flexibility to change it dynamically using ansible or other options.

net-misc/loki-promtail-bin/files/promtail.yml
```diff
@@ -0,0 +1,18 @@
+server:
+  http_listen_port: 9080
+  grpc_listen_port: 0
+
+positions:
+  filename: /tmp/positions.yaml
+
+clients:
+  - url: 'http://localhost:3100/loki/api/v1/push'
```
sibuthomasmathew, line 9: Is this a dummy value?

net-misc/loki-promtail-bin/files/promtail.yml
```diff
@@ -0,0 +1,18 @@
...
+
+scrape_configs:
+  - job_name: system
+    static_configs:
+      - targets:
+          - localhost
+        labels:
+          job: varlogs
+          __path__: /var/log/*log
```
sibuthomasmathew, line 18: Do we want promtail reading all the log files?

sibuthomasmathew (changes requested):

net-misc/loki-promtail-bin/files/promtail.yml
```diff
@@ -0,0 +1,18 @@
+server:
+  http_listen_port: 9080
+  grpc_listen_port: 0
```
sibuthomasmathew, line 3: We don't need promtail running as a server. Let's disable it
```suggestion
 disable: true
```

net-misc/loki-promtail-bin/files/promtail.yml
ibra-adjust: OK, I will remove the configuration file.

sibuthomasmathew (approved):

PR #873, opened 2024-05-02
redis: add redis 6.2.5

zegeek:
redis 6.2.5 was dropped from upstream, this avoids upgrading to another version.
https://adjustcom.atlassian.net/browse/PE-879


oziey88 (approved):
LGTM

PR #881, opened 2024-07-30
resurrect perl-5.36 to ease perl upgrades

alip:


alip:
This wont work, gentoo upgraded SLOT dependencies to 5.38 too...

PR #885, opened 2024-09-05
aerospike exporter upgraded to 1.18.0 with support for v7

leomuso:
So far tested on `esh-aero-tracker-1` only, new version available in the server.


leomuso:
Hi @a-cato, I'd like to have your comments on this before I merge and continue deployment with further servers, when you find the time of course. Thanks in advance

PR #891, opened 2024-10-09
app-admin/vault: new package

monsieurp:
https://adjustcom.atlassian.net/browse/PE-2525


app-admin/vault/vault-1.17.6.ebuild
```diff
@@ -0,0 +1,82 @@
...
+
+FILECAPS=(
+	-m 755 'cap_ipc_lock=+ep' usr/bin/${PN}
+)
+
+RESTRICT="test"
+
+src_unpack() {
+	default
+}
```
alip, line 37: This function is only calling its defaults implementation. I assume this is to prevent another eclass' unpack from overriding. If this is the case, can you please add a comment here so it'll be clear? Otherwise we can remove this function altogether.

monsieurp: function deleted

alip (approved):

PR #906, opened 2025-02-07
python-consul: bump version

licebmi:
Add supported python versions and rename tornado as the portage tree requires


dev-python/python-consul/python-consul-1.1.0-r1.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2025 Gentoo Authors
```
lagman, line 1: Genuine question, why Gentoo authors?

licebmi: https://gitweb.gentoo.org/proj/ebuild-mode.git/tree/ebuild-mode.el

PR #914, opened 2025-03-18
 keepalived_exporter_pid file and command path fix

Zia-Adjust:
https://adjustcom.atlassian.net/browse/PE-3523

### DIFF
--- a/app-metrics/keepalived_exporter/files/keepalived_exporter.initd
+++ b/app-metrics/keepalived_exporter/files/keepalived_exporter.initd
@@ -4,10 +4,11 @@
 
 name="keepalive_exporter daemon"
 description="Prometheus exporter for Keepalived metrics."
-command="/usr/bin/keepalived_exporter"
+pidfile="/run/keepalived_exporter.pid"
+command="/usr/local/bin/keepalived-exporter"
 output_log="/var/log/keepalived_exporter/keepalived_exporter.log"
 error_log="/var/log/keepalived_exporter/keepalived_exporter.err.log"
-start_stop_daemon_args="--background --user keepalived_exporter"
+start_stop_daemon_args="--background --user keepalived_exporter --make-pidfile --pidfile ${pidfile}"
 
 depend() {
     need net


app-metrics/keepalived_exporter/files/keepalived_exporter.initd
```diff
@@ -4,6 +4,7 @@
 
 name="keepalive_exporter daemon"
 description="Prometheus exporter for Keepalived metrics."
+pidfile="/run/keepalived_exporter.pid"
```
oziey88, line 7:
```suggestion
pidfile="/run/${RC_SVCNAME}.pid"
```

if it make sense to put it in dynamic way.

oziey88 (approved):

app-metrics/keepalived_exporter/files/keepalived_exporter.initd
Zia-Adjust: Yes, I can but I was looking every other name was mentioned statically in the file so I don't think so we need to do dynamic just pid name.

teerapatrk: Additional to Ozi's request, either keep `pidfile` the same, or change `pidfile`, `output_log` and `error_log` filenames together.
The latter one will be benefitcally when we rename the service daemon or need more than 1 `keepalived_exporter` daemon at a time, other than that static filename is fine for now.

Zia-Adjust:
I manually set this in `ams-shard-buffer-2` of server keepalived_exporter process and start again the process but it went to crash.

Zia-Adjust:
FTR: If I try to use this binary then pidfile didn't work and went to crash. 
```
sysops@ams-shard-buffer-2 ~ $ ls -la /usr/bin/keepalived_exporter
-rwxr-xr-x 1 root root 12004014 Feb 26 11:47 /usr/bin/keepalived_exporter

**Comment by Zia-Adjust** [comment]
I would need to update the command again because it seems like someone installed that manually
```
sysops@ams-shard-buffer-2 ~ $ qlist -v app-metrics/keepalived_exporter
app-metrics/keepalived_exporter-0.7.1 CONTENTS:
/var/log/keepalived_exporter/.keep_app-metrics_keepalived_exporter-0
/etc/conf.d/keepalived_exporter
/etc/init.d/keepalived_exporter
/usr/bin/keepalived_exporter
sysops@ams-shard-buffer-2 ~ $

Zia-Adjust:
As discussed with Sassan there will be different package will be use so we don't need to merge it.

PR #917, opened 2025-04-10
openldap: retrieve 2.4.59-r3 version for aerospike

zegeek:
aerospike debian builds are still built against debian9 which didn't still runs openldap 2.4.x . this overlay allows to install this package until we upgrade aerospike to a debian12 version where they started using openldap 2.5.x (starting from aerospike 6.4.0.0)
https://adjustcom.atlassian.net/browse/PE-3689


ibra-adjust (approved):
LGTM

PR #921, opened 2025-05-13
python-utils-r1.eclass: sync new python version support

licebmi:


teerapatrk:
@licebmi Let's sync the changes from Gentoo's upstream https://github.com/gentoo/gentoo/blob/master/eclass/python-utils-r1.eclass
@zegeek What's do you think?

teerapatrk:
Right, then I would go with sync from upstream and adjust the supported versions of Python/EAPI for backward compatibility.
With this approach, we will keep minimum number of changes tracking with the future releases.

zegeek (approved):

PR #927, opened 2025-06-26
bagger: update package size and salt

zegeek:
we changed a small part in the bagger code so it means the package resolves to a new hash and has a different size


lemanchester (approved):
LGTM

